Issue #4, opened 2019-06-05
Add Client::CONFIG_INDEX_PREFIX


renanbr:
So it would make a `Client` and its `Indexer` manage only namespaced indexes. It may be useful in cases we share the same database across multiple applications.

lyrixx:
:+1: (and also for test env)

damienalexandre:
That's not a bad idea :+1: thank you! Will look into it.

damienalexandre:
Aaanndd it's done: https://github.com/jolicode/elastically#clientconfig_index_prefix-optional

Thanks a lot!

PR #5, opened 2019-06-05
Add Client::CONFIG_INDEX_PREFIX

renanbr:
Attempt to #4

### DIFF
--- a/src/ResultSetBuilder.php
+++ b/src/ResultSetBuilder.php
@@ -54,7 +54,7 @@ class ResultSetBuilder implements BuilderInterface
 
     public function buildModelFromIndexAndData($indexName, $data)
     {
-        $pureIndexName = IndexBuilder::getPureIndexName($indexName);
+        $pureIndexName = $this->client->getIndexBuilder()->getPureIndexName($indexName);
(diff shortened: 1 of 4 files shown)


renanbr:
This PR does not check for unhandlable indexes in `Indexer` neither `IndexBuilder` (except for `purgeOldIndices()` and partially `markAsLive()`).

It's too much add an `UnhandlableIndexException`?

src/ResultSetBuilder.php
```diff
@@ -54,7 +54,7 @@ private function buildModel(Result $result)
 
     public function buildModelFromIndexAndData($indexName, $data)
     {
-        $pureIndexName = IndexBuilder::getPureIndexName($indexName);
```
damienalexandre: Hum, not a big fan of this change, because "getIndexBuilder" is kind of dumb at the moment:

https://github.com/jolicode/elastically/blob/210f52b7f2f67d19a5f8939a7cc91b60b363761b/src/Client.php#L23-L26

So it's going to boot a new instance of the builder for each search result :fearful:
Maybe we should move the `getPureIndexName` method in the Client?

PR #10, opened 2019-10-20
Add Prefixed Indexes optional feature

damienalexandre:
This PR replace https://github.com/jolicode/elastically/pull/5 from @renanbr to add a new commit:

- avoid building the IndexBuilder multiple times 
- move the "pure index name" logic
- change the prefix logic to add a `_` between the prefix and the name
- add tests and documentation

cc #4

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -16,6 +16,7 @@ class Client extends ElasticaClient
     /* Elastically config keys */
     const CONFIG_MAPPINGS_DIRECTORY = 'elastically_mappings_directory';
     const CONFIG_INDEX_CLASS_MAPPING = 'elastically_index_class_mapping';
+    const CONFIG_INDEX_PREFIX = 'elastically_index_prefix';
     const CONFIG_SERIALIZER_CONTEXT_PER_CLASS = 'elastically_serializer_context_per_class';
     const CONFIG_SERIALIZER = 'elastically_serializer';
     const CONFIG_BULK_SIZE = 'elastically_bulk_size';
@@ -38,6 +39,8 @@ class Client extends ElasticaClient
 
     public function getIndex($name): Index
     {
+        $name = $this->getConfigValue(self::CONFIG_INDEX_PREFIX).$name;
+
         return new Index($this, $name);
     }
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -49,18 +49,17 @@ class Client extends ElasticaClient
         return new Index($this, $name);
     }
 
-    public function getPrefixedIndex($pureName): string
+    public function getPrefixedIndex(string $name): string
     {
         $prefix = $this->getConfigValue(self::CONFIG_INDEX_PREFIX);
-
         if ($prefix) {
-            return sprintf('%s_%s', $prefix, $pureName);
-        } else {
-            return $pureName;
+            return sprintf('%s_%s', $prefix, $name);
         }
+
+        return $name;
     }
 
-    public function getPureIndexName($indexName): string
+    public function getPureIndexName(string $fullIndexName): string
     {
         $prefix = $this->getConfigValue(Client::CONFIG_INDEX_PREFIX);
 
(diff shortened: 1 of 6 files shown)


src/Client.php
```diff
@@ -38,9 +44,39 @@ public function getIndexer(): Indexer
...
         return new Index($this, $name);
     }
 
+    public function getPrefixedIndex($pureName): string
+    {
+        $prefix = $this->getConfigValue(self::CONFIG_INDEX_PREFIX);
+
+        if ($prefix) {
+            return sprintf('%s_%s', $prefix, $pureName);
+        } else {
```
lyrixx: No need for the `else` keyword since the `if` part has a `return`

src/Client.php
```diff
@@ -38,9 +44,39 @@ public function getIndexer(): Indexer
 
     public function getIndex($name): Index
     {
+        $name = $this->getPrefixedIndex($name);
+
         return new Index($this, $name);
     }
 
+    public function getPrefixedIndex($pureName): string
```
lyrixx: missing typehint (same for `getPureIndexName`)

src/Client.php
```diff
@@ -16,15 +16,21 @@ class Client extends ElasticaClient
...
     const CONFIG_BULK_SIZE = 'elastically_bulk_size';
 
     private $indexer;
+    private $indexBuilder;
 
     public function getIndexBuilder(): IndexBuilder
     {
-        return new IndexBuilder($this, $this->getConfig(self::CONFIG_MAPPINGS_DIRECTORY));
+        if (!$this->indexBuilder) {
+            $this->indexBuilder = new IndexBuilder($this, $this->getConfig(self::CONFIG_MAPPINGS_DIRECTORY));
```
lyrixx: This not really DI friendly :/

src/IndexBuilder.php
```diff
@@ -84,6 +80,8 @@ public function migrate(Index $current, Index $new)
 
     public function purgeOldIndices($indexName): array
```
lyrixx: missing type hint

src/Client.php
```diff
@@ -38,9 +44,39 @@ public function getIndexer(): Indexer
...
+            return $pureName;
+        }
+    }
+
+    public function getPureIndexName($indexName): string
+    {
+        $prefix = $this->getConfigValue(Client::CONFIG_INDEX_PREFIX);
+
+        if ($prefix) {
+            $pattern = sprintf('/%s_(.+)_\d{4}-\d{2}-\d{2}-\d+/i', preg_quote($prefix, '/'));
```
lyrixx: I would not hardcode the `_` here, and let it to the responsibility of the developer

src/Client.php
damienalexandre: What do you recommand?

The IndexBuiler needs the Client, the Client needs the IndexBuilder :grimacing:

src/Client.php
damienalexandre: It make it easier to force it, for regexp, and to make sure index name is always readable.

src/Client.php
lyrixx: Oh, so maybe it's a design issue. Let's keep it for now, we will see how to make it better later

src/Client.php
lyrixx: I don't like it, but OK :+1:

lyrixx: Just one question : we could use https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-meta-field.html to store more information and keep index name clean

renanbr: :+1: for `_meta`, but I think it must left some bytes to avoid collisions anyway

damienalexandre: This is a nice idea, but it would mean querying Elasticsearch for the Index (HTTP GET query). As we need the Pure Index name for each doc of a resultSet (buildModelFromIndexAndData), this is not acceptable.

PR #16, opened 2020-01-09
Add experimental support for Symfony Messenger

damienalexandre:
Symfony Messenger is really simple to use. In this pull request I just add the basic of handling indexation request via Messenger. 

There is a simple DTO, the provided handler must be overwritten to implement the fetchModel method, then I call the proper methods on the Indexer.

I use a very close version of this in production already :yum:

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ $ php bin/console messenger:consume async
 
 ### Grouping IndexationRequest in a spool
 
-Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provide a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
+Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provides a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
 
 To use this mechanism, we send the simple `IndexationRequest` in a memory transport to be consumed and grouped in a really async transport:
 
--- a/README.md
+++ b/README.md
@@ -249,9 +249,9 @@ $ php bin/console messenger:consume async
 
 ### Grouping IndexationRequest in a spool
 
-Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provides a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
+Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provides a Kernel listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
 
-To use this mechanism, we send the simple `IndexationRequest` in a memory transport to be consumed and grouped in a really async transport:
+To use this mechanism, we send the `IndexationRequest` in a memory transport to be consumed and grouped in a really async transport:
 
 ```yaml
 messenger:
@@ -264,7 +264,7 @@ messenger:
         'JoliCode\Elastically\Messenger\IndexationRequest': queuing
 ```
 
-You also need to register the event:
+You also need to register the subscriber:
 
 ```yaml
 services:
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Opinionated [Elastica](https://github.com/ruflin/Elastica) based framework to bo
 - Analysis is separated from mappings;
 - 100% compatibility with [ruflin/elastica](https://github.com/ruflin/Elastica);
 - Designed for Elasticsearch 7+ (no types), compatible with both ES 6 and ES 7;
-- Symfony Messenger Handler support;
+- Symfony Messenger Handler support (with or without spool);
 - Extra commands to monitor, update mapping, reindex... Commonly implemented tasks.
 
 ## Demo
@@ -173,7 +173,96 @@ Add a prefix to all indexes and aliases created via Elastically.
 
 _Default to `null`._
 
+## Usage in Symfony
+
+### Client as a service
+
+Just declare the proper service in `services.yaml`:
+
+```yaml
+JoliCode\Elastically\Client:
+    arguments:
+        $config:
+            log: '%kernel.debug%'
+            host: '%env(ELASTICSEARCH_HOST)%'
+            elastically_mappings_directory: '%kernel.root_dir%/Elasticsearch/mappings'
+            elastically_index_class_mapping:
+                my_index_name: \App\Model\MyModel
+            elastically_bulk_size: 100
+```
+
+### Using Messenger for async indexing
+
+Elastically ship a default Message and Handler for Symfony Messenger.
+
+Register the message in your configuration:
+
+```yaml
+framework:
+    messenger:
+        transports:
+            async: "%env(MESSENGER_TRANSPORT_DSN)%"
+
+        routing:
+            # async is whatever name you gave your transport above
+            'JoliCode\Elastically\Messenger\IndexationRequest':  async
+
+services:
+    App\Messenger\IndexationRequestHandler: ~
+```
+
+You have to implement `App\Messenger\IndexationRequestHandler` by extending the core `\JoliCode\Elastically\Messenger\IndexationRequestHandler` so you can plug your database or any other source of truth.
+
+Then from your code you have to call:
+
+```php
+use JoliCode\Elastically\Messenger\IndexationRequest;
+use JoliCode\Elastically\Messenger\IndexationRequestHandler;
+
+$bus->dispatch(new IndexationRequest(Product::class, '1234567890'));
+
+// Third argument is the operation, so for a delete:
+// new IndexationRequest(Product::class, 'ref9999', IndexationRequestHandler::OP_DELETE);
+```
+
+And then consume the messages:
+
+```sh
+$ php bin/console messenger:consume async
+```
+
+### Grouping IndexationRequest in a spool
+
+Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provide a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
+
+To use this mechanism, we send the simple `IndexationRequest` in a memory transport to be consumed and grouped in a really async transport:
+
+```yaml
+messenger:
+    transports:
+        async: "%env(MESSENGER_TRANSPORT_DSN)%"
+        queuing: 'in-memory:///'
+
+    routing:
+        'JoliCode\Elastically\Messenger\MultipleIndexationRequest': async
+        'JoliCode\Elastically\Messenger\IndexationRequest': queuing
+```
+
+You also need to register the event:
+
+```yaml
+services:
+    JoliCode\Elastically\Messenger\IndexationRequestSpoolSubscriber:
+        arguments:
+            - '@messenger.transport.queuing' # should be the name of the memory transport
+            - '@messenger.default_bus'
+        tags:
+            - { name: kernel.event_subscriber }
+```
+
 ## Using Jane for DTO and fast Normalizers
+
+:warning: See https://github.com/jolicode/elastically/issues/12 before.
 
 Install [JanePHP](https://jane.readthedocs.io/) and the model generator to build your own DTO and Normalizers. Then create your Serializer like this:
 
@@ -192,24 +281,6 @@ $client = new Client([
     Client::CONFIG_SERIALIZER => $serializer,
 ]);
 ```
-
-## Usage in Symfony
-
-Just declare the proper service in `services.yaml`:
-
-```yaml
-JoliCode\Elastically\Client:
-    arguments:
-        $config:
-            log: '%kernel.debug%'
-            host: '%env(ELASTICSEARCH_HOST)%'
-            elastically_mappings_directory: '%kernel.root_dir%/Elasticsearch/mappings'
-            elastically_index_class_mapping:
-                my_index_name: \App\Model\MyModel
-            elastically_bulk_size: 100
-```
-
-> todo, document the Messenger Handler usage.
 
 ## To be done
 
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ JoliCode\Elastically\Client:
             host: '%env(ELASTICSEARCH_HOST)%'
             elastically_mappings_directory: '%kernel.root_dir%/Elasticsearch/mappings'
             elastically_index_class_mapping:
-                my_index_name: \App\Model\MyModel
+                my_index_name: App\Model\MyModel
             elastically_bulk_size: 100
 ```
 
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ JoliCode\Elastically\Client:
 
 ### Using Messenger for async indexing
 
-Elastically ship a default Message and Handler for Symfony Messenger.
+Elastically ships with a default Message and Handler for Symfony Messenger.
 
 Register the message in your configuration:
 
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Opinionated [Elastica](https://github.com/ruflin/Elastica) based framework to bo
 - Analysis is separated from mappings;
 - 100% compatibility with [ruflin/elastica](https://github.com/ruflin/Elastica);
 - Designed for Elasticsearch 7+ (no types), compatible with both ES 6 and ES 7;
+- Symfony Messenger Handler support;
 - Extra commands to monitor, update mapping, reindex... Commonly implemented tasks.
 
 ## Demo
@@ -207,6 +208,8 @@ JoliCode\Elastically\Client:
                 my_index_name: \App\Model\MyModel
             elastically_bulk_size: 100
 ```
+
+> todo, document the Messenger Handler usage.
 
 ## To be done
 
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.14",
         "phpunit/phpunit": "^8",
-        "symfony/messenger": "^4.1|^5.0"
+        "symfony/messenger": "^4.3|^5.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,9 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.14",
         "phpunit/phpunit": "^8",
-        "symfony/messenger": "^4.3|^5.0"
+        "symfony/messenger": "^4.3|^5.0",
+        "symfony/framework-bundle": "^4.3|^5.0",
+        "symfony/http-foundation": "^4.3|^5.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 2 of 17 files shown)


composer.json
```diff
@@ -25,7 +25,8 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.14",
-        "phpunit/phpunit": "^8"
+        "phpunit/phpunit": "^8",
+        "symfony/messenger": "^4.1|^5.0"
```
lyrixx: 4.1 is not supported anymore. Let's bump this to 4.3

src/Messenger/IndexationRequest.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace JoliCode\Elastically\Messenger;
+
+final class IndexationRequest
+{
+    private $op;
```
lyrixx: don't be lazy :D 
Could you rename it to `operation` (and everywhere else?) It will enchance the DX

src/Messenger/IndexationRequest.php
```diff
@@ -0,0 +1,36 @@
...
+    private $type;
+    private $id;
+
+    public function __construct(string $type, string $id, string $op = IndexationRequestHandler::OP_INDEX)
+    {
+        if (!in_array($op, IndexationRequestHandler::OPS, true)) {
```
lyrixx: Could you run php-cs-fixer to automatically add `\` in front of all PHP functions that are opitmised by opcache

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+
+use Elastica\Document;
+use JoliCode\Elastically\Client;
+use Psr\Log\NullLogger;
+use Symfony\Component\Messenger\Exception\UnrecoverableMessageHandlingException;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+abstract class IndexationRequestHandler implements MessageHandlerInterface
+{
+    const OP_INDEX = 'index';
```
lyrixx: should we add the visibility to all constants?

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+    const OP_DELETE = 'delete';
+    const OP_UPDATE = 'update';
+    const OP_CREATE = 'create';
+
```
lyrixx: There is no support for bulk?

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+                break;
+        }
+
+        $indexer->flush();
+    }
+
+    /**
+     * Return a model (DTO) to send for indexation.
+     */
+    abstract public function fetchModel(string $type, string $id);
```
lyrixx: Should be before `__invoke`, isn't ?
PHP CS Fixer will tell :)

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+    ];
+
+    private $client;
+
+    public function __construct(Client $client)
+    {
+        $this->client = $client;
+
+        // Disable the logs for memory concerns
+        $this->client->setLogger(new NullLogger());
```
lyrixx: This is very opinated and in a consumer it will remove it forever.
Another move would be to wrap the content of `__invoke()` in `a try {} finally {}` to remove it only during the execution of this handler

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public function __invoke(IndexationRequest $message)
+    {
+        $indexer = $this->client->getIndexer();
+        $indexToClass = $this->client->getConfig(Client::CONFIG_INDEX_CLASS_MAPPING);
+        $indexName = array_search($message->getType(), $indexToClass, true);
+
+        if (!$indexName) {
+            throw new UnrecoverableMessageHandlingException(sprintf('The given type (%s) does not exist!', $message->getType()));
```
lyrixx: I would remplace the `!` by a `.`

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        if (self::OP_DELETE === $message->getOp()) {
+            $indexer->scheduleDelete($indexName, $message->getId());
+            $indexer->flush();
```
lyrixx: What if there is alredy something in the buffer? 
I would not flush it here

src/Messenger/IndexationRequest.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace JoliCode\Elastically\Messenger;
+
+final class IndexationRequest
+{
+    private $op;
+    private $type;
+    private $id;
+
+    public function __construct(string $type, string $id, string $op = IndexationRequestHandler::OP_INDEX)
```
lyrixx: Why do you call `type` like that? It's actually an index, isn't ?
CF: the doc:
```php
// New Client object with new options
$client = new Client([
    Client::CONFIG_INDEX_CLASS_MAPPING => [
        'beers' => Beer::class,    
    ],
])
```

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,81 @@
...
+
+            return;
+        }
+
+        $model = $this->fetchModel($message->getType(), $message->getId());
+
+        if (!$model) {
+            // ID does not exists, delete
```
lyrixx: Why?

src/Messenger/IndexationRequest.php
damienalexandre: Named it that way because the property is serialized and stored in a bus. So `operation` is 7 bytes heavier... I will rename it for the sake of DX :+1:

src/Messenger/IndexationRequest.php
damienalexandre: To be done in another PR, I already runned `make cs` but adding `\` is not part of our configuration.

src/Messenger/IndexationRequestHandler.php
damienalexandre: Not at the moment. I'm not sure how best to do it - didn't need such support in other projects.

src/Messenger/IndexationRequestHandler.php
damienalexandre: PHP CS Fixer didn't care :yum:

src/Messenger/IndexationRequestHandler.php
damienalexandre: If not here, where should we persist our changes?

src/Messenger/IndexationRequest.php
damienalexandre: It's the class name we expect here, not the index name.

src/Messenger/IndexationRequestHandler.php
damienalexandre: To me, if `fetchModel` fail to found the entity/the data, that means it does not exists. If it did exists before, triggering a delete is a safe way to sync the `fetchModel` behavior with what we have in Elasticsearch.

src/Messenger/IndexationRequestHandler.php
lyrixx: In the bits of code that call the handler. I apply the same advice with doctrine. I only flush in controller, command, consumer and everything is simpler like that

src/Messenger/IndexationRequest.php
lyrixx: so classname is better? As you may see, I didn't understand this code :/

src/Messenger/IndexationRequestHandler.php
lyrixx: But if the does not exist, if you delete it you basically delete nothing.

src/Messenger/IndexationRequestHandler.php
renanbr: In a project I work on, we're introducing bulking operations in a messenger consumer for elasticsearch sync.

Not yet finished, but the gain is around 65% in terms of time.

If we adopt elastically (my planing) it would be great not get blocked in this subject. I mean I'm for "non flush" + "bulk"

damienalexandre: For "bulk"; it means being able to apply the same operation on multiple ID's. That's a good suggestion I will change the Messenger message to support this.

But I don't get the "non flush" part: if I don't flush the Indexer, updates will be lost. @lyrixx made the same suggestion, I have difficulties understanding how it can work. Do you suggest being able to run:

```php
$this->bus->dispatch(new IndexationRequest(....));
$this->indexer->flush();
```

??

src/Messenger/IndexationRequestHandler.php
damienalexandre: The logger consume *a lot* of memory, specially in a worker mode.

Another way could be to "clear" the logger at each `__invoke`... I don't understand what a try catch could solve?

lyrixx: It's not a try catch, it's a try finally :) 
It's an easy way to do something, and undo it whatever happen in the try (IE: the finally is always executed)

src/Messenger/IndexationRequestHandler.php
lyrixx: If the handler is called by the workers command (so async), there are no other place to call async.
I'm gonna mark my comment as resolved

src/Messenger/IndexationRequestHandler.php
lyrixx: as stated in another comment, there are no issue with the flush (since the handler can be used in a async transport)

src/Messenger/IndexationRequestHandler.php
damienalexandre: This will have to be handled in another PR, created https://github.com/jolicode/elastically/issues/17 to reference this.

damienalexandre:
I just pushed a better version and the documentation.

There this a "MultipleIndexationRequest" message now, and a listener to group single messages in this one, allowing spool / bulk of all the changes into one message!

src/Messenger/IndexationRequestHandler.php
damienalexandre: Support for Bulk is added!

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,94 @@
...
+            $this->schedule($indexer, $message);
+        }
+
+        $indexer->flush();
+    }
+
+    private function schedule(Indexer $indexer, IndexationRequest $indexationRequest)
+    {
+        $indexToClass = $this->client->getConfig(Client::CONFIG_INDEX_CLASS_MAPPING);
+        $indexName = array_search($indexationRequest->getClassName(), $indexToClass, true);
```
damienalexandre: Found a bug here. 

I suggest to make this resolver part of the Client.

damienalexandre:
The pooling feature is tested successfully on a real world project :tada:

src/Messenger/IndexationRequestSpoolSubscriber.php
```diff
@@ -0,0 +1,71 @@
...
+        $this->singleTransport = $singleTransport;
+        $this->bus = $bus;
+    }
+
+    public function onException()
+    {
+        $this->wasExceptionThrown = true;
+    }
+
+    public function onTerminate()
```
renanbr: Isn't expected to check `$this->wasExceptionThrown` here?

damienalexandre: Absolutely! Fixed, thanks!

src/Messenger/IndexationRequestSpoolSubscriber.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $message = new MultipleIndexationRequest($operations);
+        $this->bus->dispatch($message);
+    }
+
+    public static function getSubscribedEvents()
+    {
+        $listeners = [
+            KernelEvents::EXCEPTION => 'onException',
+            KernelEvents::TERMINATE => 'onTerminate',
```
lyrixx: I would listen kernel.response instead of this event

damienalexandre: It's better for debug yes, 
but it's not really intended for that IMO.

> This event allows you to modify or replace the response that will be replied.

Any other opinions about that?

README.md
```diff
@@ -224,6 +207,97 @@ JoliCode\Elastically\Client:
             ...
 ```
 
+### Using Messenger for async indexing
+
+Elastically ship a default Message and Handler for Symfony Messenger.
```
pyrech:
```suggestion
Elastically ships with a default Message and Handler for Symfony Messenger.
```

README.md
```diff
@@ -224,6 +207,97 @@ JoliCode\Elastically\Client:
...
+
+And then consume the messages:
+
+```sh
+$ php bin/console messenger:consume async
+```
+
+### Grouping IndexationRequest in a spool
+
+Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provide a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
```
pyrech:
```suggestion
Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provides a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
```

src/Messenger/IndexationRequestSpoolSubscriber.php
lyrixx: We don't really care :) it's just a generic description.
There is an event, we can leverage it 
If you prefer, you can use `kernel.finish_request`, but it's almost the same things...

README.md
```diff
@@ -224,6 +207,97 @@ JoliCode\Elastically\Client:
...
+
+And then consume the messages:
+
+```sh
+$ php bin/console messenger:consume async
+```
+
+### Grouping IndexationRequest in a spool
+
+Sending multiple `IndexationRequest` during the same Symfony Request is not always appropriate, it will trigger multiple Bulk operations. Elastically provides a Kernel terminate listener to group all the `IndexationRequest` in a single `MultipleIndexationRequest` message.
```
lyrixx: 1. it's not a kernel terminate listener. I would simply remove `terminate` 
2. Why naming it `Multiple` and not `Bulk`

README.md
```diff
@@ -224,6 +207,97 @@ JoliCode\Elastically\Client:
...
+
+To use this mechanism, we send the simple `IndexationRequest` in a memory transport to be consumed and grouped in a really async transport:
```
lyrixx:
```suggestion
To use this mechanism, we send the `IndexationRequest` in a memory transport to be consumed and grouped in a really async transport:
```

README.md
```diff
@@ -224,6 +207,97 @@ JoliCode\Elastically\Client:
...
+    transports:
+        async: "%env(MESSENGER_TRANSPORT_DSN)%"
+        queuing: 'in-memory:///'
+
+    routing:
+        'JoliCode\Elastically\Messenger\MultipleIndexationRequest': async
+        'JoliCode\Elastically\Messenger\IndexationRequest': queuing
+```
+
+You also need to register the event:
```
lyrixx:
```suggestion
You also need to register the subscriber:
```

composer.json
```diff
@@ -26,7 +26,10 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.14",
         "phpunit/phpunit": "^8",
-        "symfony/http-client": "^4.3|^5.0"
+        "symfony/http-client": "^4.3|^5.0",
+        "symfony/messenger": "^4.4|^5.0",
+        "symfony/framework-bundle": "^4.3|^5.0",
+        "symfony/http-foundation": "^4.3|^5.0"
```
lyrixx: 1. Could we sort packages by name?
2. and add the composer config option to automatically sort packages

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        $indexer = $this->client->getIndexer();
+
+        if ($message instanceof MultipleIndexationRequest) {
+            foreach ($message->getOperations() as $operation) {
+                $this->schedule($indexer, $operation);
+            }
+        }
+
+        if ($message instanceof IndexationRequest) {
```
lyrixx: It could be an `elseif` ?

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,94 @@
...
+            case self::OP_UPDATE:
+                $indexer->scheduleUpdate($indexName, new Document($indexationRequest->getId(), $model, '_doc'));
+                break;
+        }
+    }
+
+    /**
+     * Return a model (DTO) to send for indexation.
+     */
+    abstract public function fetchModel(string $className, string $id);
```
lyrixx: 1. You may add the return type `Document` (and so remove the PHP Doc
2. It could be abstract

src/Messenger/IndexationRequestSpoolSubscriber.php
```diff
@@ -0,0 +1,85 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\Messenger\MessageBusInterface;
+use Symfony\Component\Messenger\Transport\TransportInterface;
+
+/**
+ * Same idea as https://github.com/symfony/swiftmailer-bundle/blob/master/EventListener/EmailSenderListener.php.
+ */
+class IndexationRequestSpoolSubscriber implements EventSubscriberInterface
```
lyrixx: It could implements `ResetInterface` (and so add the tag on the service)

src/Messenger/MultipleIndexationRequest.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace JoliCode\Elastically\Messenger;
+
+final class MultipleIndexationRequest implements IndexationRequestInterface
+{
+    private $operations = [];
+
+    public function __construct($operations)
```
lyrixx:
```suggestion
    public function __construct(array $operations)
```

README.md
damienalexandre: Naming it `Bulk` refer to the low level Elastica implementation, but at this level, it's just "multiple indexation request", you don't know if it's one, two, three bulk requests...

src/Messenger/IndexationRequestHandler.php
damienalexandre: I see... Naming it "fetchDocument" instead. Why not :thinking: Will look into that.

src/Messenger/IndexationRequestHandler.php
```diff
@@ -0,0 +1,92 @@
...
+
+        if ($message instanceof MultipleIndexationRequest) {
+            foreach ($message->getOperations() as $operation) {
+                $this->schedule($indexer, $operation);
+            }
+        } elseif ($message instanceof IndexationRequest) {
+            $this->schedule($indexer, $message);
+        }
+
+        $indexer->flush();
```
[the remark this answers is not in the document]
damienalexandre: With Messenger, any Exception is gonna re-queue the entire message.

In the case of Elasticsearch indexing, a bulk can fail only for one doc, and the other docs will be OK in the index (no rollback). I like the idea but it means we should edit the message that's being "re-queue"; I don't know how to do that yet :)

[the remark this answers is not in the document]
renanbr: I tried to implement it here: https://github.com/jolicode/elastically/pull/18

PR #18, opened 2020-02-01
Requeue only failed messages when handling MultipleIndexationRequest

renanbr:
This is an attempt to implement https://github.com/jolicode/elastically/pull/16#discussion_r370675531

### DIFF
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -137,4 +137,9 @@ class Indexer
             $this->flush();
         }
     }
+
+    public function getBulkMaxSize(): int
+    {
+        return $this->bulkMaxSize;
+    }
 }
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -137,9 +137,4 @@ class Indexer
             $this->flush();
         }
     }
-
-    public function getBulkMaxSize(): int
-    {
-        return $this->bulkMaxSize;
-    }
 }
--- a/src/Messenger/IndexationRequestHandler.php
+++ b/src/Messenger/IndexationRequestHandler.php
@@ -49,22 +49,31 @@ abstract class IndexationRequestHandler implements MessageHandlerInterface
         }
 
         $indexer = $this->client->getIndexer();
-        $initialQueueSize = $indexer->getQueueSize();
+
+        $messageOffset = 0;
+        $responseOffset = $indexer->getQueueSize();
 
         try {
             foreach ($messages as $indexationRequest) {
+                ++$messageOffset;
                 $this->schedule($indexer, $indexationRequest);
+
+                if (0 === $indexer->getQueueSize()) {
+                    $responseOffset = 0;
+                }
             }
 
             $indexer->flush();
         } catch (ResponseException $exception) {
             // Extracts failed operations from the bulk
+            // Responses are checked in reverse mode because we have only requests from the last bulk
             $failedMessages = [];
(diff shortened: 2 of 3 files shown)


src/Messenger/IndexationRequestHandler.php
```diff
@@ -26,27 +28,58 @@ abstract class IndexationRequestHandler implements MessageHandlerInterface
...
+            $messages = [$message];
         }
 
-        $indexer->flush();
+        $indexer = $this->client->getIndexer();
+        $originalBulkMaxSize = $indexer->getBulkMaxSize();
+        $initialQueueSize = $indexer->getQueueSize();
+
+        try {
+            $indexer->setBulkMaxSize($initialQueueSize + count($messages));
```
damienalexandre: I'm not a big fan of this. 

Messenger Handler can be used in a variety of setup including sync, and this will change the Indexer behavior application-wide. Also it drops the threshold, the Max Bulk Size is necessary to avoid sending huge payload over the network (what if I have 6000 messages in my MultipleIndexationRequest?).

It can be done differently;
let the Indexer "flush" whenever it wants, and you will still get an Exception.

When that occurs, we know where we are in the foreach loop:
- what messages have been sent (and could be in error)
- what messages are not sent and can be processed next

WDYT?

renanbr: I start working in this in b673c446498c419a50e46a72037a73615319e1c0

Anyway it doesn't cover yet the case a non-first bulk triggers an exception

renanbr: finished via e8b0918d122c3533e0ba9d064b813d95a22eb103

src/Messenger/IndexationRequestHandler.php
```diff
@@ -26,27 +28,67 @@ abstract class IndexationRequestHandler implements MessageHandlerInterface
...
+            }
+
+            // Throws exception as-is if all operations have failed
+            if (count($failedMessages) === count($messages)) {
+                throw $exception;
+            }
+
+            // Redispatch failed or non-executed messages
+            $nonExecutedMessages = array_slice($messages, $messageOffset);
+            $toRedispatch = array_merge($failedMessages, $nonExecutedMessages);
```
renanbr: We may want to dispatch `$nonExecutedMessages` as a single `MultipleIndexationRequest` instead

renanbr: done via 5a20a5e16b862bae83f4e79d8e56fb4502c488b6

src/Indexer.php
```diff
@@ -137,4 +137,9 @@ public function setBulkMaxSize(int $bulkMaxSize): void
             $this->flush();
         }
     }
+
+    public function getBulkMaxSize(): int
+    {
+        return $this->bulkMaxSize;
+    }
```
damienalexandre: Can you remove this getter? It's not used anymore :+1:

renanbr: done via fef2f6ed8d283f6120c0e5229a7fc057405ff0f9

damienalexandre:
Looks like your test case disappeared :fearful:

renanbr:
> Looks like your test case disappeared

I messed it up during rebase :sweat_smile: but I found thanks to `reflog`, done via a773c4ff6a33ca1831120ac450d4b3be870fddb9

damienalexandre:
Thanks a lot :+1:

PR #19, opened 2020-02-10
bump elastica version to ^7

nicolassing:


src/Client.php
```diff
@@ -42,7 +43,11 @@ public function getIndexer(): Indexer
         return $this->indexer;
     }
 
-    public function getIndex($name): Index
+    /**
+     * @param string $name
+     * @return ElasticaIndex|Index
```
damienalexandre: This phpDoc is wrong, the code is very explicit here, we return an Elastically Index object, 
could you tell us more about this change?

src/IndexBuilder.php
```diff
@@ -74,8 +72,10 @@ public function speedUpRefresh(Index $index): void
 
     public function migrate(Index $current, Index $new)
     {
-        // @todo Waiting for https://github.com/ruflin/Elastica/pull/1637 to be merged
-        // This method should use the TASK API, because we do not want to WAIT for the reindex (HTTP Timeout issues).
+        $reindex = new Reindex($current, $new);
+        $reindex->setWaitForCompletion(Reindex::WAIT_FOR_COMPLETION_FALSE);
+
+        return $reindex->run();
```
damienalexandre: As stated in the original comment, I would like this method to wait (or allow to wait).
But we can't wait on the original HTTP call for timeout reasons. So we should use the task ID and the Task API to monitor this long running process.

Could you put back the comment maybe?

tests/IndexBuilderTest.php
```diff
@@ -169,4 +171,56 @@ public function testPurgeAndCloseOldIndices(): void
...
+
+        $client = $this->getClient();
+        $document = $client->getIndex($indexName)->getDocument('f');
+        $this->assertInstanceOf(Document::class, $document);
+
+        $indexBuilder = $this->getIndexBuilder(__DIR__.'/configs');
+        $newIndex = $indexBuilder->createIndex('empty');
+
+        $indexBuilder->migrate($client->getIndex($indexName), $newIndex);
+        usleep(1200000); // 1,2 second
```
damienalexandre: This is kind of random, but we can keep this while waiting for the "pooling task API" feature, cf my other comment.

src/Client.php
lyrixx: Could we add PHP Type hint everywhere instead?

src/IndexBuilder.php
```diff
@@ -74,8 +72,10 @@ public function speedUpRefresh(Index $index): void
 
     public function migrate(Index $current, Index $new)
     {
-        // @todo Waiting for https://github.com/ruflin/Elastica/pull/1637 to be merged
-        // This method should use the TASK API, because we do not want to WAIT for the reindex (HTTP Timeout issues).
+        $reindex = new Reindex($current, $new);
```
renanbr: It would be great if this library provide a way to configure how `reindex` will run, I mean choosing a strategy to slice, and to specify a pipeline, or even to run multiples reindex with specific queries each one

... unless the goal is to provide a basic migrate strategy

imHo, this is out of scope of "bump elastica version to ^7", and could be place in another PR

nicolassing: Originally I made this PR to use this lib with Elasticsearch 7.3 and so I had to use Elastica ^7. For now it's working well and I start to see some improvement I could make. 
Actually I don't have time to work on it but I would in the next months :)

BTW Great work ! <3

damienalexandre: Thanks for your feedback, I think we will take over your changes in separated pull requests, so we can merge the Elastica version bump soon :+1:

damienalexandre:
@nicolassing Hi Nicolas. Elastic 7 is out and I took a spin at the upgrade here #29. 

I'm not closing your PR yet because there is also a reindex implementation we may want to add. I will try to reuse your commit in another PR.

damienalexandre:
- Elastica 7 is done in #29 
- Reindex is done in #30 (I manager to keep ownership of your commit :muscle:)

I'm closing this PR, thank a lot!

Issue #20, opened 2020-02-21
HttpClientTransport doesn't pass the expected bulk payload


renanbr:
When I try to bulk operations in a client that uses `HttpClientTransport` as transport I get this error:

```
Elastica\Exception\ResponseException: The bulk request must be terminated by a newline [\n]
```

damienalexandre:
Fixed on master

PR #21, opened 2020-02-21
Add test for bulk over symfony/http-client transport

renanbr:
This PR tries to illustrate the issue reported at #20 

Note that I've targeted it to `feat/message` because this branch has some important changes in how tests and dependencies are managed.


damienalexandre:
Thanks! Good catch :+1:

damienalexandre:
Just fixed on master :+1: Thanks again!

Issue #22, opened 2020-03-08
Get Index/Mapping definition through annotation / database / whatever


ismail1432:
Hello :wave: 
Thanks for this library :+1: 

There is few months I had to work on a solution to define mapping and instead of using yaml I used annotations, do you think it's relevant to add a workaound this to this lib ?

```php
use App\Search\Annotation as Els;

/**
 * @Els\Index(
 *     name="product",
 *     settings={
 *         "analysis": {
 *          "number_of_replicas": 1,
 *          "refresh_interval": 60s,
 *             "analyser": {
 *                 "bpm_analyser": {
 *                     "tokenizer": "standard"
 *                 }
 *             }
 *         }
 *     })
 */
class ProductDocument

 /**
  * @var string
  *
  * @Els\ScalarMapping(type="string") // Default value
  */
 public $subtitle;

 /**
  * @var bool
  *
  * @Els\ScalarMapping(type="boolean")
  */
 public $storage = false;
```

damienalexandre:
Hello,
Thanks for the suggestion! 

I'm not a big fan of annotation in the library directly for some reasons:

- you have to learn new syntax / naming for Elasticsearch stuff (ScalarMapping?)
- if a new option came to Elasticsearch, and is not exposed as annotation, developers are stuck / limited to what is implemented
- it can be very verbose / hard to read, if you want multiple "fiels", etc...

**But** that should be made possible. I already had questions about having another source of truth for the configuration (I have clients storing this in a database i.e.).

The **IndexBuilder** class could rely on a new abstract **Configuration** provider, which is YAML by default and can be switched by whatever your project needs. 

Would you be able to work on such change?

ismail1432:
Thanks for the quick reply,

Even annotation seems not a good solution I think you're right and we should let developer have the choice of the source.

> The IndexBuilder class could rely on a new abstract Configuration provider[..] Would you be able to work on such change?

damienalexandre:
The possibility to fetch the configuration freely is also needed here: https://github.com/jolicode/elastically/pull/50

The File adapter should allow this king of custom naming.

jmsche:
As stated in #62 it would be nice to at least have PHP as an alternative language to configure analyzers & mapping.

My current case is indexing several types of Formations which have a same base "tree" but some properties differ.

I'd also like to be able to configure it using attributes if someone works on annotations as well :)

lyrixx:
see #79

lyrixx:
[...]

PR #23, opened 2020-03-15
[POC] Add possibility to switch configuration 

ismail1432:
Following https://github.com/jolicode/elastically/issues/22 I poc a simple way to switch configuration, by default we always take configuration from yaml file but we can also define a custom configuration class. 

```yaml
    JoliCode\Elastically\Client:
        arguments:
            $config:
                configuration_class: '@App\Search\DudeConfiguration' 
```

```diff
+  configuration_class: '@App\Search\DudeConfiguration' 
```

`App\Search\DudeConfiguration` can be define as a service so you can inject any others service needed. Suppose that I need to fecth configuration from a database

```php
class DudeConfiguration extends AbstractConfiguration
{
    public function __construct(array $options = [], EntityManagerInterface $manager)
    {
        $config = [
            'elastically_mappings_directory' => $manager->getRepository(Config::class)->findOneBy(['configName'=> 'path']),
            'elastically_bulk_size' => $manager->getRepository(Config::class)->findOneBy(['configName'=>'bulkSize'])
            ];
   
        parent::__construct($config);
    }
}
```
At the moment we not handle the mapping definition but 
I think we can add a option `elastically_mapping` that return the whole mapping and in `IndexBuilder::createIndex` we get it the mapping from the option resolver. 
Something like :

```diff
- $mappingFilePath = $this->configuration->getOption('elastically_mappings_directory').DIRECTORY_SEPARATOR.$indexName.'_mapping.yaml';
- $mapping = Yaml::parseFile($mappingFilePath);
+ $mapping = $this->configuration->getOption('elastically_mappings')
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -3,8 +3,11 @@
 namespace JoliCode\Elastically;
 
 use Elastica\Client as ElasticaClient;
+use Elastica\Exception\RuntimeException;
+use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
 use Symfony\Component\Serializer\Encoder\JsonEncoder;
 use Symfony\Component\Serializer\Normalizer\ArrayDenormalizer;
+use Symfony\Component\Serializer\Normalizer\DateTimeNormalizer;
 use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
 use Symfony\Component\Serializer\Serializer;
 use Symfony\Component\Serializer\SerializerInterface;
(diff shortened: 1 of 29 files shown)


src/Client.php
```diff
@@ -25,10 +25,13 @@ class Client extends ElasticaClient
     private $indexer;
     private $indexBuilder;
 
+    /** @var AbstractConfiguration $configuration */
+    private $configuration;
```
damienalexandre, line 29: I thinks what you wanted to achieve here is more a "MappingReader" than a "Configuration"; configuration is still an array given to the Client and this new AbstractConfiguration class looks ambiguous.

src/Client.php
```diff
@@ -25,10 +25,13 @@ class Client extends ElasticaClient
...
+
     public function getIndexBuilder(): IndexBuilder
     {
         if (!$this->indexBuilder) {
-            $this->indexBuilder = new IndexBuilder($this, $this->getConfig(self::CONFIG_MAPPINGS_DIRECTORY));
+            $this->indexBuilder = new IndexBuilder($this, $this->getConfiguration());
```
damienalexandre, line 34: Maybe having "mapping reader" is better for understanding the code here?

```suggestion
            $this->indexBuilder = new IndexBuilder($this, $this->getMappingReader());
```

src/Client.php
ismail1432: I do agree I updated in this way

src/Client.php
ismail1432: fixed

ismail1432:
> Some comments about naming things :)
> 
> Also could you fix the tests?
> 
> Thanks a lot

Thanks for the feedback, I updated the code with your comments,
and I also replace the logic to resolve the configuration to get them from mapping reader **that still be from the yaml by default** 
Exemple :

```diff
- $indexToClass = $this->getConfig(Client::CONFIG_INDEX_CLASS_MAPPING);
+ $indexToClass = $this->getMappingReader()->getOption(self::CONFIG_INDEX_CLASS_MAPPING);
```

I'm still open if there is something wrong or anything to improve and OFC I'll fix test :+1:

ismail1432:
Hi @damienalexandre,

Sorry for the delay, I'll close this one as I went too far

PR #25, opened 2020-04-03
Check candidate index name suffix before purge it

### LINKED ISSUE #24: Purge method is too permissive
I got an issue where the wrong indices have been deleted/purged.

If you have two indices:

- foo
- foo_bar

When you call `$this->indexBuilder->purgeOldIndices('foo');` it will delete `foo_bar`.

It should only delete indices called `foo_[DATE]`. We must check for a date pattern and valide the result.

renanbr:
This PR is an attempt to fix #24

With this fix the `IndexBuilder->purgeOldIndices()` will skip purging indices which their suffix isn't a valid date time.

---

To think about: Should it relays only on the index name? Why not track the "public" name and created date in the index metadata?

### DIFF
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -97,7 +97,14 @@ class IndexBuilder
                 continue;
             }
 
-            $date = \DateTime::createFromFormat('Y-m-d-His', str_replace($indexName.'_', '', $realIndexName));
+            // Check suffix (it must contains a valid date)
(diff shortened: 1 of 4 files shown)


damienalexandre:
Yes we could use index metadata to store the "real" name of an index, that way we can run the purge only on indexes that have this metadata.

Your fix seems good too, I don't know exactly if adding a layer of complexity is a good idea, isn't the name + date check enough?

damienalexandre:
Let's merge "as is", it's a good fix! thanks

PR #27, opened 2020-05-06
Add tests for Jane JSON Schema 6.x, allow fast DTO In and Out

### LINKED ISSUE #12: Compatibility with Jane is not possible without hacky code
Elastica always reads the JSON from Elasticsearch inside an associative array (and we cannot change that without rewriting a lot of code).

So when it's time for Elastically to hydrate the DTO, we only have the array version of our "_source".

This is not an issue on the default installation of Elastically but when using Jane as the DTO + Normalizer source, it breaks.

Jane Normalizer expects `stdObject`, not arrays.

The quick and dirty hack to fix this is to run:

```
$source = json_decode(json_encode($source), false);
```

Before: https://github.com/jolicode/elastically/blob/bcb348f7ef25ac4a7dfdb627bdffb4d6127fce19/src/ResultSetBuilder.php#L65

How can we fix this in an elegant and robust way?

cc @Korbeil @joelwurtz @lyrixx 

Error:

> PHP Fatal error:  Uncaught Symfony\Component\Serializer\Exception\InvalidArgumentException: Given $data is not an object (array given). We need an object in order to continue denormalize method.

damienalexandre:
Jane 6.0 fixes #12, which is great news.

This PR adds tests for Jane support by building a schema and using the DTO for Indexing and Search.

### DIFF
--- a/tests/Jane/JaneTest.php
+++ b/tests/Jane/JaneTest.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace JoliCode\Elastically\Tests\Jane;
+
+use Elastica\Document;
+use Jane\JsonSchema\Console\Command\GenerateCommand;
+use Jane\JsonSchema\Console\Loader\ConfigLoader;
+use Jane\JsonSchema\Console\Loader\SchemaLoader;
+use JoliCode\Elastically\Client;
+use JoliCode\Elastically\Tests\Jane\generated\Model\MyModel;
+use JoliCode\Elastically\Tests\Jane\generated\Model\MyModelIngredientsItemAnyOf;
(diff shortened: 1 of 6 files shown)


tests/Jane/JaneTest.php
```diff
@@ -0,0 +1,91 @@
...
+        ], $command->getDefinition());
+
+        $command->execute($inputArray, new NullOutput());
+
+        // Build the Serializer
+        $normalizers = [
+            new ArrayDenormalizer(),
+            new JaneObjectNormalizer(),
+        ];
+        $encoders = [new JsonEncoder(
```
Korbeil, line 41: You only need to do `new JsonEncoder()` to have the same result

damienalexandre:
I took the setup from https://jane.readthedocs.io/en/latest/documentation/JsonSchema.html#using ^^

Also the tests dont use Flex so I need to build everything here.

PR #29, opened 2020-06-23
Upgrade Elastica to 7.x and bump PHP to 7.4

damienalexandre:
Implement the newly released Elastica 7.x.

~~I had to bump PHP to 7.4 because we need https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters.~~


lyrixx:
> I had to bump PHP to 7.4 because we need https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters.

I did not see where you needed that. Is it already implemented?

src/Client.php
```diff
@@ -43,7 +43,7 @@ public function getIndexer(): Indexer
         return $this->indexer;
     }
 
-    public function getIndex($name): Index
+    public function getIndex(string $name): Index
```
damienalexandre, line 46: We need PHP 7.4 here exactly @lyrixx.

The parent class type hint with `\Elastica\Index`, and we need `\JoliCode\Elastically\Index`. It's just one line but I can't get around it.

damienalexandre:
As this PR may be the last before I tag version 1.0 I will tag a new alpha right now, before merging. Because we may have projets relying on dev-master that will not be able to install this version.

PR #35, opened 2020-07-15
Use kernel.project_dir instead of kernel.root_dir in example

rogamoore:
kernel.root_dir is deprecated since Symfony 4.2

see https://symfony.com/blog/new-in-symfony-4-2-important-deprecations

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ JoliCode\Elastically\Client:
     arguments:
         $config:
             host: '%env(ELASTICSEARCH_HOST)%'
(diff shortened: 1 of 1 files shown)


damienalexandre:
Awesome, thanks again! :+1:

damienalexandre:
It was right in the tests (https://github.com/jolicode/elastically/blob/ef33908a6f1a96f4930c75cf7d951628fba0dd74/tests/Symfony/config.yaml#L43) but wrong in the README :facepalm:

PR #40, opened 2020-10-13
Fix issue while dealing with prefixed Index.

atapatel:
Fix issue while dealing with prefixed Index. When i try to use prefix index for elastic search purge of indexing is not working. It is taking two time prefixing index name while closing and deleting indexes.

### DIFF
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -145,14 +145,14 @@ class IndexBuilder
                 $livePassed = true;
             }
 
-             if ($livePassed && $afterLiveCounter > 1) {
+            if ($livePassed && $afterLiveCounter > 1) {
                 // Remove
-                $realIndexName = str_replace($this->client->getPrefixedIndex(''),'',$realIndexName);
+                $realIndexName = str_replace($this->client->getPrefixedIndex(''), '', $realIndexName);
                 $this->client->getIndex($this->client->getPureIndexName($realIndexName))->delete();
                 $operations[] = sprintf('%s deleted.', $realIndexName);
             } elseif ($livePassed && 1 === $afterLiveCounter) {
                 // Close
-                $realIndexName = str_replace($this->client->getPrefixedIndex(''),'',$realIndexName);
+                $realIndexName = str_replace($this->client->getPrefixedIndex(''), '', $realIndexName);
                 $this->client->getIndex($this->client->getPureIndexName($realIndexName))->close();
                 $operations[] = sprintf('%s closed.', $realIndexName);
             }
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -153,7 +153,7 @@ class IndexBuilder
             } elseif ($livePassed && 1 === $afterLiveCounter) {
                 // Close
                 $index = new Index($this->client, $realIndexName);
-                $index->delete();
+                $index->close();
                 $operations[] = sprintf('%s closed.', $realIndexName);
             }
         }
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -145,13 +145,15 @@ class IndexBuilder
                 $livePassed = true;
             }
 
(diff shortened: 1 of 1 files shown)


damienalexandre:
Are you calling purgeOldIndices with the real index name? (without the prefix I mean).

atapatel:
@damienalexandre  Yes, i am using this library along with this library https://github.com/monsieurbiz/SyliusSearchPlugin.

atapatel:
@damienalexandre Am I doing something wrong here?

damienalexandre:
Yes this change looks strange to me, and the tests are red: https://travis-ci.org/github/jolicode/elastically/jobs/735342582#L448

The purgeOldIndices function should not care about the prefix option because it fetch the real concrete name from the State API. The test show how it's supposed to be used:

https://github.com/jolicode/elastically/blob/223752b125f17e7ccb8563adc0657c1eccfa3a94/tests/IndexBuilderTest.php#L134-L172

Here, "hop" is my index name, which is used as an alias in Elasticsearch. The concrete index name will be something like "PREFIX_INDEXNAME_DATE". We need to add a test with a prefix configured to check the behavior. Could you try? (running the tests is done via `make test`.

_Also, the CS is broken, can you run `make cs`?_

atapatel:
@damienalexandre Do you want me to create test cases for this scenario?

damienalexandre:
Yes that would be great, or wait until end of the week, I will have some time to look into this issue as well.

Thank you!

src/IndexBuilder.php
```diff
@@ -147,11 +147,13 @@ public function purgeOldIndices(string $indexName): array
 
             if ($livePassed && $afterLiveCounter > 1) {
                 // Remove
-                $this->client->getIndex($realIndexName)->delete();
+                $realIndexName = str_replace($this->client->getPrefixedIndex(''), '', $realIndexName);
+                $this->client->getIndex($this->client->getPureIndexName($realIndexName))->delete();
```
damienalexandre, line 151:
```suggestion
                $index = new Index($this->client, $realIndexName);
                $index->delete();
```

src/IndexBuilder.php
```diff
@@ -147,11 +147,13 @@ public function purgeOldIndices(string $indexName): array
...
                 $operations[] = sprintf('%s deleted.', $realIndexName);
             } elseif ($livePassed && 1 === $afterLiveCounter) {
                 // Close
-                $this->client->getIndex($realIndexName)->close();
+                $realIndexName = str_replace($this->client->getPrefixedIndex(''), '', $realIndexName);
+                $this->client->getIndex($this->client->getPureIndexName($realIndexName))->close();
```
damienalexandre, line 156:
```suggestion
                $index = new Index($this->client, $realIndexName);
                $index->close();
```

damienalexandre (changes requested):
Hello!
I just added a test for this here: https://github.com/jolicode/elastically/commit/2a67fcee93fab0d78226b3d4778d869280c8b2ae

You were right there is a bug when using the purge method with prefixed indices.

I suggest to change the code that effectively delete and closes the indexes to avoid any "name" manipulation. As we have the real index name, but can't use $client->getIndex(), creating the Elastica Index object ourself is the better way to access it for real.

What do you think?

src/IndexBuilder.php
atapatel: This looks good :)

src/IndexBuilder.php
atapatel: :+1:

atapatel:
I will update the code according to your suggestion and push code here. :) . Thanks a lot, @damienalexandre . You have created a really nice library.

atapatel:
@damienalexandre Just one suggestion can you label this pull request as `Hacktoberfest` and `good first issue` ?

damienalexandre:
All done, thanks!

PR #42, opened 2020-10-15
Throw an exception when the Indexer queue is not cleared

damienalexandre:
This is kind of experimental but I had this idea where we can prevent developers from misusing the Indexer.

Upon __destruct, we check the Indexer queue, and if there is Documents here we throw an exception :boom:.

My only fear is that:

> Attempting to throw an exception from a destructor (called in the time of script termination) causes a fatal error.

Is it an issue?

This PR also adds a "clear" method to empty the queue.

### DIFF
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -18,11 +18,16 @@ class Indexer
 
     public function __construct(Client $client, SerializerInterface $serializer, int $bulkMaxSize = 100)
     {
-        // TODO: on the destruct, maybe throw an exception for non empty indexer queues?
-
         $this->client = $client;
         $this->bulkMaxSize = $bulkMaxSize ?? 100;
         $this->serializer = $serializer;
+    }
+
+    public function __destruct()
+    {
+        if ($this->getQueueSize() > 0) {
+            throw new \RuntimeException(sprintf('%s queue is not empty, you need to call flush() to send documents to Elasticsearch or clear() to empty the queue.', __CLASS__));
+        }
     }
 
     public function scheduleIndex($index, Document $document)
@@ -102,6 +107,11 @@ class Indexer
         $this->currentBulk = null;
 
         return $response;
+    }
+
+    public function clear(): void
+    {
+        $this->currentBulk = null;
     }
 
     public function getQueueSize()
--- a/tests/IndexerTest.php
+++ b/tests/IndexerTest.php
@@ -39,6 +39,30 @@ final class IndexerTest extends BaseTestCase
         $this->assertEquals('f', $document->getId());
     }
 
+    public function testIndexDocumentsWithoutFlushingThrowError(): void
+    {
+        $indexName = mb_strtolower(__FUNCTION__);
+        $dto = new TestDTO();
+        $dto->bar = 'Oops I forgot the flush.';
+
+        $indexer = $this->getIndexer();
+
+        $indexer->scheduleIndex($indexName, new Document('f', $dto));
+        $this->assertEquals(1, $indexer->getQueueSize());
+        $indexer->clear();
+        $this->assertEquals(0, $indexer->getQueueSize());
+
+        $indexer->scheduleIndex($indexName, new Document('f', $dto));
+        try {
+            $indexer->__destruct();
+            $this->assertFalse(true, 'This code should not be accessible.');
+        } catch (\RuntimeException $exception) {
+            $this->assertStringContainsString('queue is not empty', $exception->getMessage());
+        }
+
+        $indexer->clear(); // avoid fatal error when the test ends
+    }
+
     public function testIndexOneDocumentWithMapping(): void
     {
         $indexName = mb_strtolower(__FUNCTION__);


jdreesen:
Why not flush the queue on destruct, instead of throwing an exception?

damienalexandre:
Yeah that's also possible. I'm open to suggestions :+1:

renanbr:
Hello you all,

I'm not sure requesting a database or throwing an exception in the destructor is a good thing.

IMO people should rely on framework triggers to handle it. As this lib is Symfony friendly, we could add an event listener to do something in the `kernel.finish_request` or `kernel.terminate` for example.

lyrixx (approved):
👍🏼  I like it

jdreesen:
> I'm not sure **requesting a database** or throwing an exception in the destructor is a good thing.

You're right. I'm doing this in my current project (which uses another elasticsearch libaray):

```php
    public function __destruct()
    {
        if ($this->bulkCount > 0) {
            $this->commit();
        }
    }
```

It mostly (99% of the time) runs fine, but sometimes it throws like this at the end:

```
Fatal error: Uncaught ErrorException: Warning: curl_setopt_array(): supplied resource is not a valid cURL handle resource in /var/www/html/vendor/guzzlehttp/ringphp/src/Client/CurlFactory.php on line 52

Call Stack:
    0.0002     392512   1. {main}() /var/www/html/bin/console:0
    8.6368   83180968   2. MyApp\Console\Application->run() /var/www/html/bin/console:26

ErrorException: Warning: curl_setopt_array(): supplied resource is not a valid cURL handle resource in /var/www/html/vendor/guzzlehttp/ringphp/src/Client/CurlFactory.php on line 52

Call Stack:
    0.0002     392512   1. {main}() /var/www/html/bin/console:0
    8.6368   83180968   2. MyApp\Console\Application->run() /var/www/html/bin/console:26
   21.3915  103627104   3. AppBundle\Search\DocumentRepository\DocumentRepository->__destruct() /var/www/html/src/AppBundle/Search/DocumentRepository/DocumentRepository.php:0
   21.3915  103627104   4. AppBundle\Search\DocumentRepository\DocumentRepository->commit() /var/www/html/src/AppBundle/Search/DocumentRepository/DocumentRepository.php:29
   21.3915  103627104   5. ONGR\ElasticsearchBundle\Service\IndexService->commit() /var/www/html/src/AppBundle/Search/DocumentRepository/DocumentRepository.php:64
   21.3915  103627480   6. Elasticsearch\Client->bulk() /var/www/html/vendor/ongr/elasticsearch-bundle/Service/IndexService.php:455
   21.3918  103640088   7. Elasticsearch\Client->performRequest() /var/www/html/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Client.php:792
   21.3918  103640128   8. Elasticsearch\Transport->performRequest() /var/www/html/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Client.php:1586
   21.3918  103640128   9. Elasticsearch\Connections\Connection->performRequest() /var/www/html/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php:107
   21.3918  103641256  10. Elasticsearch\Connections\Connection->Elasticsearch\Connections\{closure:/var/www/html/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php:212-318}() /var/www/html/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php:187
   21.3918  103632776  11. GuzzleHttp\Ring\Client\Middleware::GuzzleHttp\Ring\Client\{closure:/var/www/html/vendor/guzzlehttp/ringphp/src/Client/Middleware.php:28-32}() /var/www/html/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php:218
   21.3918  103632776  12. GuzzleHttp\Ring\Client\CurlHandler->__invoke() /var/www/html/vendor/guzzlehttp/ringphp/src/Client/Middleware.php:30
   21.3918  103632872  13. GuzzleHttp\Ring\Client\CurlHandler->_invokeAsArray() /var/www/html/vendor/guzzlehttp/ringphp/src/Client/CurlHandler.php:68
   21.3918  103632872  14. GuzzleHttp\Ring\Client\CurlFactory->__invoke() /var/www/html/vendor/guzzlehttp/ringphp/src/Client/CurlHandler.php:84
   21.3919  103635696  15. curl_setopt_array() /var/www/html/vendor/guzzlehttp/ringphp/src/Client/CurlFactory.php:52
   21.3919  103636744  16. Symfony\Component\Debug\ErrorHandler->handleError() /var/www/html/vendor/guzzlehttp/ringphp/src/Client/CurlFactory.php:52
```

So it's probably not a good idea to do this...

jdreesen:
> Why not flush the queue on destruct, instead of throwing an exception?

Instead of flushing on `__destruct()` we could flush via `register_shutdown_function()`, too. I'm not very familiar with these functions, I just stumbled upon it and thought it may be an option.

lyrixx:
Thanks for exploring this! After looking at it again with the current codebase, I don't think throwing from `__destruct()` is the right approach:

1. **It ends in a fatal error in real usage.** The `Indexer` is a shared service (the Symfony container, `Factory::buildIndexer()`), so it's usually destroyed at script shutdown. At that point, a thrown exception can't be caught and just becomes a fatal error. The only case where it behaves nicely is the explicit `__destruct()` call in the test.
2. **It can hide the real error.** In `IndexationRequestHandler`, if one operation of a `MultipleIndexationRequest` throws something other than `ResponseException` (for example an `UnrecoverableMessageHandlingException` when no index is mapped), the documents already scheduled stay in the queue. The destructor exception would then replace the original one.
3. **Tests become unpredictable.** A test failing between `schedule*()` and `flush()` would throw later, during GC, possibly in another test.

Flushing in `__destruct()` has the same problem: I/O in destructors is fragile, as the cURL stack trace above shows.

The `clear()` method is worth keeping, though. Point 2 is also a bug on its own: the leftover operations get sent with the *next* message handled by the worker. I'd suggest:
- a small PR adding `Indexer::clear()` and calling it in `IndexationRequestHandler` when handling fails;
- for the "forgotten flush" problem, a Symfony listener (`kernel.terminate` / `console.terminate` / `WorkerMessageHandledEvent`) that flushes or logs a warning, and/or implementing `ResetInterface`.

WDYT? Could we close this one in favor of these?

damienalexandre:
I think I like the second solution, just an error log triggered by the proper Symfony listeners :+1:

PR #47, opened 2020-10-22
Code refactor.

atapatel:
- Refactore flush logic which is commonly used to sync operation with Elastic search in batch.


damienalexandre:
Thanks @atapatel

PR #50, opened 2020-11-12
Add optional fileName parameter to specify an index file name

JochenRousse:
Hi, we have an use case where we'd like to specify one mapping for multiple indexes (one index per locale).
This change would allow this use case without breaking existing code.

### DIFF
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -24,8 +24,8 @@ class IndexBuilder
 
     public function createIndex($indexName, $fileName = null): Index
     {
-        $fileName = $fileName ?? $indexName;
-        $mappingFilePath = $this->configurationDirectory.DIRECTORY_SEPARATOR.$fileName.'_mapping.yaml';
+        $fileName = $fileName ?? ($indexName.'_mapping.yaml');
+        $mappingFilePath = $this->configurationDirectory.DIRECTORY_SEPARATOR.$fileName;
         if (!is_file($mappingFilePath)) {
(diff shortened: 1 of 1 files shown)


src/IndexBuilder.php
```diff
@@ -22,9 +22,10 @@ public function __construct(Client $client, $configurationDirectory)
         $this->configurationDirectory = $configurationDirectory;
     }
 
-    public function createIndex($indexName): Index
+    public function createIndex($indexName, $fileName = null): Index
     {
-        $mappingFilePath = $this->configurationDirectory.DIRECTORY_SEPARATOR.$indexName.'_mapping.yaml';
+        $fileName = $fileName ?? $indexName;
+        $mappingFilePath = $this->configurationDirectory.DIRECTORY_SEPARATOR.$fileName.'_mapping.yaml';
```
Korbeil, line 28:
```suggestion
        $fileName = $fileName ?? ($indexName.'_mapping.yaml');
        $mappingFilePath = $this->configurationDirectory.DIRECTORY_SEPARATOR.$fileName;
```
Why not this version ? In order to allow more possible fileName and remove that `_mapping.yaml` part.

JochenRousse: Yes indeed, thanks :)

damienalexandre:
Thanks a lot!

Just as mention, I'm planning to refactor the Mapping Provider for version 2 => https://github.com/jolicode/elastically/issues/22

PR #57, opened 2021-02-08
Fix getIndexNameFromClass with index prefix

scuben:
`getIndexNameFromClass` now also working when using an index-prefix configuration.


damienalexandre:
Thanks a lot!

Red tests are not related :+1:

PR #59, opened 2021-02-15
Add support for PHP 8.0

jmsche:


src/IndexBuilder.php
```diff
@@ -130,7 +130,7 @@ public function purgeOldIndices(string $indexName): array
 
         // Newest first
         uasort($indexes, function ($a, $b) {
-            return $a['date'] < $b['date'];
+            return $a['date'] < $b['date'] ? 1 : -1;
```
jdreesen, line 133: Why not using the spaceship operator?
```suggestion
            return $b['date'] <=> $a['date'];
```

jmsche: Thanks, I always forget about this one :)

composer.json
```diff
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.2|^8.0",
```
Korbeil, line 18: Single pipe is not recommended and deprecated, you should use double pipe here.
```suggestion
        "php": "^7.2 || ^8.0",
```

It was in documentation before, but can't find where, so here is an issue about this https://github.com/composer/composer/issues/6755

jmsche: TIL, thank you @Korbeil :)

lyrixx (approved):

Korbeil (approved):

damienalexandre:
Thanks a lot everyone!

PR #61, opened 2021-03-04
Enhance bulk counting

francisbesset:
Avoid to use `\count()` on each action, just increment and read a variable for better performance.

### DIFF
--- a/src/Bulk.php
+++ b/src/Bulk.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace JoliCode\Elastically;
+
+use Elastica\Bulk\Action;
+use Elastica\Bulk as ElasticaBulk;
+
+class Bulk extends ElasticaBulk
+{
+    protected $size = 0;
+
+    public function getSize()
+    {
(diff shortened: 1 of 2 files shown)


lyrixx:
Did you notice a real impact on performance?

damienalexandre:
Hi!

Thanks for your suggestion. I must say introducing a new class on top of Elastica is a big deal and it must be done only if highly necessary. 

In the case of `count` I'm really curious about possible performance impact your contribution brings, do you have any numbers to share?

Thanks!

francisbesset:
> Did you notice a real impact on performance?

With a quickly phpbench, count() function is 5-10% slower than an increment var.

lyrixx:
I mean, in a real project 😁
It looks like a premature optimisation

lyrixx:
Hello @francisbesset 

Thanks for your patch, but as I said in my previous comment, I don't think this optimisation is really useful.

That's why I'm closing this PR.

Anyway, thanks for the work :)

PR #65, opened 2021-07-02
Expose the raw elastica result in serializer context while denormalizing

lyrixx:


src/ResultSetBuilder.php
```diff
@@ -35,34 +28,34 @@ private function buildResults(Response $response): array
...
+        $source = $result->getSource();
+        if (!$source) {
+            return null;
         }
 
-        $context = $this->client->getSerializerContext($indexToClass[$pureIndexName]);
+        $class = $this->client->getClassFromIndexName($this->client->getPureIndexName($result->getIndex()));
+
+        $context = $this->client->getSerializerContext($class);
+        $context['result'] = $result;
```
damienalexandre, line 57: Maybe add a prefix or something? `elastically_result`? To avoid any side effect on existing context.

src/Client.php
```diff
@@ -77,6 +78,17 @@ public function getIndexNameFromClass(string $className): string
...
+    {
+        $indexToClass = $this->getConfig(self::CONFIG_INDEX_CLASS_MAPPING);
+
+        if (!isset($indexToClass[$indexName])) {
+            throw new RuntimeException(sprintf('Unknown class for index "%s", did you forgot to configure "%s"?', $indexName, self::CONFIG_INDEX_CLASS_MAPPING));
+        }
+
+        return $indexToClass[$indexName];
+    }
+
```
damienalexandre, line 91: :+1:

damienalexandre (changes requested):
This introduce inconsistency between search results and "get" results :thinking: 
https://github.com/jolicode/elastically/blob/56824ea0ea870615cebb881e8dcb83befc46ab07/src/Index.php#L17

So if your serializer needs the "result" context it cannot work on non-search results - this will have to be used responsibly.

src/ResultSetBuilder.php
Korbeil: & add a const in the class to define the context key (like the Serializer does for example) if possible

lyrixx:
> So if your serializer needs the "result" context it cannot work on non-search results - this will have to be used responsibly.

yes, of course 👍🏼 but we can not do anything for that 🙈

src/ResultSetBuilder.php
lyrixx: thanks, fixed

Korbeil (approved):

src/ResultSetBuilder.php
```diff
@@ -27,51 +30,52 @@ public function __construct(Client $client)
...
-
-    private function buildResults(Response $response): array
     {
         $data = $response->getData();
         $results = [];
 
         if (!isset($data['hits']['hits'])) {
-            return $results;
+            return new ResultSet($response, $query, []);
         }
```
damienalexandre, line 39: Not sure this should be changed as it will break the compatibility with Elastica: https://github.com/ruflin/Elastica/blob/cbc962e22060cf7b1bafd403cf3027943e20ae15/src/ResultSet/DefaultBuilder.php#L32-L34

damienalexandre (commented):
Except for my comment it looks good :+1:

src/ResultSetBuilder.php
lyrixx: It's the same behavior than before :) 
I inline the method, that's why there is this diff 🌻

damienalexandre: No, before it returned an array, now it return a ResultSet.

lyrixx: Double no :)

damienalexandre: ok I see, my bad.

damienalexandre (approved):

PR #66, opened 2021-07-02
Added PHP-CS-Fixer config + run it

lyrixx:
I took the config from https://github.com/jolicode/starfleet/blob/master/.php_cs

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,43 @@
+<?php
+
+$header = <<<'EOF'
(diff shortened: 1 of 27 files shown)


damienalexandre (changes requested):
Thanks, can you add a check in the test suite?

lyrixx:
@damienalexandre Not really :/ The CI is dead, we should migrate to GA, so let's add later with GA?

damienalexandre (approved):

PR #72, opened 2021-07-21
Drop support for PHP < 7.4

lyrixx:


src/IndexBuilder.php
```diff
@@ -22,16 +22,16 @@
...
 
-    public function __construct(Client $client, $configurationDirectory)
+    public function __construct(Client $client, string $configurationDirectory)
     {
         $this->client = $client;
         $this->configurationDirectory = $configurationDirectory;
     }
 
-    public function createIndex($indexName, $fileName = null): Index
+    public function createIndex(string $indexName, ?string $fileName = null): Index
```
jdreesen, line 34: Note that adding parameter types is technically a breaking change, because it forces users extending this class to add them, too.

damienalexandre:
You also need to update the `README` (the PHP version is mentioned)

lyrixx:
> You also need to update the `README` (the PHP version is mentioned)

Done ✅

PR #76, opened 2021-07-21
Switch to Github Action instead of travis

lyrixx:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+            image: docker.elastic.co/elasticsearch/elasticsearch-oss:7.8.0
+            ports:
+              - '9200:9200'
+        steps:
+            - name: Wait for ES to be ready
+              run: |
+                curl 127.0.0.1:9200 || true
+                curl localhost:9200 || true
+                docker ps
+                wget -q --waitretry=1 --retry-connrefused -T 10 -O - http://127.0.0.1:9200
```
damienalexandre, line 47: We need a green cluster, not only a responding one.

Maybe you could use what was used in travis?

https://github.com/jolicode/elastically/blob/446d90fda74bc86ece9dcec50733f2131a204682/.travis.yml#L23-L26

lyrixx: ✅

PR #78, opened 2021-07-21
Clean how the project deal with DI

lyrixx:
TODO
* [x] Externalize all "factory" to a proper class
* [x] The client use the Factory now, but all methods are deprecated
* [x] Introduce IndexNameMapper to map an index to a class (and deal with "pure" index name)
* [x] Introduce ContextBuilderInterface to retrive to build a context for the serializer
* [x] Break many inderections, and try to reduce the dependency on the Client
* [x] Fixed tests
* [x] Update README
* [x] CHANGELOG


damienalexandre (commented):
Looks good. Not a fan of all the constants being in the Factory class, if this class is only supposed to help peoples that don't use DIC? Feels likes unnecessary move, could you elaborate on the motivations behind this change?

The original code made sure to never override Elastica `__construct` methods, to be as much compatible as possible and compatible with future versions of Elastica. Do you think that was unnecessary precaution?

I'm not able to test it right now, but I'm looking forward the README change to see the impact ^^

Thanks a lot!

lyrixx:
> Looks good. Not a fan of all the constants being in the Factory class, if this class is only supposed to help peoples that don't use DIC? Feels likes unnecessary move, could you elaborate on the motivations behind this change?

To me it's odd to define constants in a class, and use theme in another class (less context switch). More over, only one `use` statement is needed with all in the same place. And finally, Why would we set these constant in the Client? It has nothing to do with the client.

> The original code made sure to never override Elastica `__construct` methods, to be as much compatible as possible and compatible with future versions of Elastica. Do you think that was unnecessary precaution?

I agree with you. I'm not very confortable with that either. But since elastically **extends** elastica, we should not avoid to override constructor : 1/ it blocks us for future developments 2/ we write more complexe code. So I override constructor, and I pretty confident because 1/ semver exists, and I hope elastica respect it 2/ we have tests (that are run each week whatever happen).

> I'm not able to test it right now, but I'm looking forward the README change to see the impact ^^

I'll do that right now :)

README.md
```diff
@@ -26,41 +28,45 @@ composer require jolicode/elastically
...
 }
 
-// Building the Index from a mapping config
-use JoliCode\Elastically\Client;
+use JoliCode\Elastically\Factory;
 use Elastica\Document;
 
-// New Client object with new options
-$client = new Client([
+// New Factory object with new options
```
damienalexandre, line 46: The logic is way different here. Before it was just an Elastica client compatible client.
Now we have to build a Factory, not a Client.

```suggestion
// Factory object with Elastica options + new Elastically options in the same array
```

README.md
```diff
@@ -26,41 +28,45 @@ composer require jolicode/elastically
...
-    Client::CONFIG_MAPPINGS_DIRECTORY => __DIR__.'/mappings',
+    Factory::CONFIG_MAPPINGS_DIRECTORY => __DIR__.'/mappings',
     // What object to find in each index
-    Client::CONFIG_INDEX_CLASS_MAPPING => [
+    Factory::CONFIG_INDEX_CLASS_MAPPING => [
         'beers' => Beer::class,
     ],
 ]);
 
+// Class to  the client
```
damienalexandre:
```suggestion
// Custom Elastica compatible Client
```

README.md
```diff
@@ -26,41 +28,45 @@ composer require jolicode/elastically
...
+    Factory::CONFIG_MAPPINGS_DIRECTORY => __DIR__.'/mappings',
     // What object to find in each index
-    Client::CONFIG_INDEX_CLASS_MAPPING => [
+    Factory::CONFIG_INDEX_CLASS_MAPPING => [
         'beers' => Beer::class,
     ],
 ]);
 
+// Class to perform request
+$client = $factory->buildClient();
```
damienalexandre, line 57:
```suggestion
// Class to perform request, same as the Elastica Client
$client = $factory->buildClient();
```

README.md
```diff
@@ -157,33 +163,48 @@ An array of index name to class FQN.
 ]
 ```
 
-### Client::CONFIG_SERIALIZER (optional)
+### `Factory::CONFIG_SERIALIZER` (optional)
 
-A `SerializerInterface` and `DenormalizerInterface` compatible object that will by used both on indexation and search.
+A `SerializerInterface` compatible object that will by used both on indexation.
```
damienalexandre, line 168:
```suggestion
A `SerializerInterface` compatible object that will by used on indexation.
```

README.md
```diff
@@ -157,33 +163,48 @@ An array of index name to class FQN.
...
 
-_Default to Symfony Object Normalizer._
+_Default to Symfony Serializer with Object Normalizer._
 
 A faster alternative is to use Jane to generate plain PHP Normalizer, see below. Also we recommend [customization to handle things like Date](https://symfony.com/doc/current/components/serializer.html#normalizers).
 
-### Client::CONFIG_SERIALIZER_CONTEXT_PER_CLASS (optional)
+### `Factory::CONFIG_DENORMALIZER` (optional)
+
+A `DenormalizerInterface` compatible object that will by used both on search.
```
damienalexandre, line 176:
```suggestion
A `DenormalizerInterface` compatible object that will by used on search results to build your objects back.
```

README.md
```diff
@@ -157,33 +163,48 @@ An array of index name to class FQN.
...
+
+If this option is not defined, the factory will fallback to
+`Factory::CONFIG_SERIALIZER` option.
+
+### `Factory::CONFIG_SERIALIZER_CONTEXT_BUILDER` (optional)
+
+An instance of `ContextBuilderInterface` that build a serializer context from a
+class name.
+
+If it is not defined, elastically, will use a `StaticContextBuilder` with the
```
damienalexandre, line 186:
```suggestion
If it is not defined, Elastically, will use a `StaticContextBuilder` with the
```

src/Factory.php
```diff
@@ -0,0 +1,123 @@
...
+            null,
+            null,
+            $this->buildBuilder(),
+            $this->buildIndexNameMapper()
+        );
+    }
+
+    public function buildIndexNameMapper(): IndexNameMapper
+    {
+        return $this->indexNameMapper ??= $this->config[self::CONFIG_INDEX_NAME_MAPPER] ?? new IndexNameMapper($this->config[self::CONFIG_INDEX_PREFIX] ?? null, $this->config[self::CONFIG_INDEX_CLASS_MAPPING] ?? []);
```
damienalexandre, line 68: I know the null coalescing operator is nice and everything, but the readability of this method does not benefit from it IMO :grimacing:

src/Client.php
```diff
@@ -12,57 +12,40 @@
...
-        return $this->indexBuilder;
-    }
-
-    public function getIndexer(): Indexer
-    {
-        if (!isset($this->indexer)) {
-            $this->indexer = new Indexer($this, $this->getSerializer(), $this->getConfigValue(self::CONFIG_BULK_SIZE, 100));
+        // BC Layer, to remove in 2.0
+        $config[Factory::CONFIG_CLIENT] = $this;
+        $this->factory = new Factory($config);
```
damienalexandre, line 39: Instead of using the `$config` array to push a Client instance, why not use a setter or a real property constructor on the Factory? To avoid injected the Client inside the Client.

lyrixx: Good idea! ✅

lyrixx:
The PR is ready ✅

damienalexandre (approved):
Nice, thanks for keeping the BC :+1:

PR #79, opened 2021-07-22
Introduce MappingProviderInterface

lyrixx:


damienalexandre (approved):

src/IndexBuilder.php
```diff
@@ -11,43 +11,31 @@
...
+    public function __construct(MappingProviderInterface $mappingProvider, Client $client, IndexNameMapper $indexNameMapper)
     {
+        $this->mappingProvider = $mappingProvider;
         $this->client = $client;
-        $this->configurationDirectory = $configurationDirectory;
         $this->indexNameMapper = $indexNameMapper;
     }
 
-    public function createIndex(string $indexName, ?string $fileName = null): Index
+    public function createIndex(string $indexName, array $context = []): Index
```
vvasiloi, line 36: This is usually a BC Break and is not done in minor version release.

damienalexandre: Yes, sorry about that. We had a lengthy internal discussion about bumping a major version for this and decided against it. We are going to be more careful next time :+1:

PR #83, opened 2021-07-23
Fixes to php-cs-fixer: ignore cache file in Git, fix Makefile usage & apply to…

jmsche:
Hi,

This PR aims to fix the following:

- includes .php-cs-fixer.php to the files to fix (& apply fixes to it)
- ignore the Jane generated files in the tests directory
- ignores php-cs-fixer cache file in Git
- fix Makefile usage to make it consistent with CI & config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /tests/Jane/generated/*
 !/tests/Jane/generated/.gitkeep
 .php_cs.cache
+.php-cs-fixer.cache
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the jolicode/elastically library.
+ *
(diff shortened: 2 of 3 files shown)


damienalexandre:
Perfect, thank you :+1:

.gitignore
```diff
@@ -4,3 +4,4 @@
 /tests/Jane/generated/*
 !/tests/Jane/generated/.gitkeep
 .php_cs.cache
```
jdreesen, line 6: This should not be needed anymore, as it's the name of the cache file from PHP-CS-Fixer 2 (see: [upgrade docs](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/UPGRADE-v3.md#rename-of-files))

lyrixx: Would you mind sending a PR? :)

jdreesen: Here it is: #85

PR #84, opened 2021-07-23
phpstan: allow to run locally & create base config

jmsche:
Hi,

I noticed I couldn't run phpstan locally & it would only run in CI, so:

- added phpstan to dev requirements
- created config to share between CI & local runs
- added phpstan in Makefile


damienalexandre:
Thanks!

PR #85, opened 2021-07-23
Remove legacy PHP-CS-Fixer cache file from .gitignore

jdreesen:
This should not be needed anymore, as it's the name of the cache file from PHP-CS-Fixer 2 (see: [upgrade docs](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/UPGRADE-v3.md#rename-of-files))


lyrixx (approved):

lyrixx:
Thanks 👍🏼

PR #86, opened 2021-07-23
Add a bridge : A bundle for Symfony

lyrixx:
```yaml
elastically:
    connections:
        default:
            client:
                host: '%env(ELASTICSEARCH_HOST)%'
            mapping_directory: '%kernel.project_dir%/config/elasticsearch'
            index_class_mapping:
                beers: App\Dto\Beer
```

See the readme for more information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Drop support support for Symfony 4.*
 - Drop support for PHP 7.4
+- Add a bundle to ease integration with Symfony
 
 ## [1.5.4] - 2022-12-16
 
--- a/README.md
+++ b/README.md
@@ -227,57 +227,99 @@ _Default to `null`._
 
 ## Usage in Symfony
 
-### Client as a service
+### Configuration
 
-Just declare the proper service in `services.yaml`:
+You'll need to add the bundle in `bundles.php`:
 
-```yaml
-services:
-    JoliCode\Elastically\IndexNameMapper:
-        arguments:
-            $prefix: null # or a string to prefix index name
-            $indexClassMapping:
-                indexName: My\AwesomeDTO
-
-    JoliCode\Elastically\Serializer\StaticContextBuilder:
-        arguments:
-            $mapping:
-                My\AwesomeDTO: []
-
-    JoliCode\Elastically\ResultSetBuilder:
-        arguments:
-            $indexNameMapper: '@JoliCode\Elastically\IndexNameMapper'
-            $contextBuilder: '@JoliCode\Elastically\Serializer\StaticContextBuilder'
-            $denormalizer: '@serializer'
-
-    JoliCode\Elastically\Client:
-        arguments:
-            $config:
-                host: '%env(ELASTICSEARCH_HOST)%'
-            $logger: '@logger'
-            $resultSetBuilder: '@JoliCode\Elastically\ResultSetBuilder'
-            $indexNameMapper: '@JoliCode\Elastically\IndexNameMapper'
-
-    JoliCode\Elastically\Indexer:
-        arguments:
-            $client: '@JoliCode\Elastically\Client'
-            $serializer: '@serializer'
-            $bulkMaxSize: 100
-            $bulkRequestParams: []
-            $contextBuilder: '@JoliCode\Elastically\Serializer\StaticContextBuilder'
-
-    JoliCode\Elastically\Mapping\YamlProvider:
-        arguments:
-            $configurationDirectory: '%kernel.project_dir%/config/elasticsearch'
-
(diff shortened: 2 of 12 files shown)


lyrixx:
I need some time...

And we need to take a decission: 
Do we support only one connection, or do we allow many?

Korbeil:
IMO, we should go simple as first implementation and support only one connection, if someone has needs for more we can create a new issue and make a new iteration of that feature later.

lyrixx:
I would like to avoid that.

Because supporting theses two configuration is **really** boring


```yaml
client:
    host: 127.0.0.1
    mapping: data/mapping.yaml

```

```yaml
client:
    default:
        host: 127.0.0.1
        mapping: data/mapping.yaml
    another:
        host: 127.0.0.2
        mapping: data/mapping2.yaml

```

I already have a POC on my computer to support N connexion. I need not so much time to finish it I guess

damienalexandre:
I agree with @lyrixx, multiple connexion from the start is good. We must solve the autowiring that's all.

lyrixx:
I just pushed a new version. **still in WIP**

The configuration has the following shape:

```
elastically:
    connections:
        default:
            prefix:               null
            index_class_mapping:
                my-index:            My\Dto
            serializer:
                context_mapping:
                    foo:                 bar
            client:               []
            mapping_directory:    ~
```

* You can have as much as client as you want.
* If there is only one client, this is the default one (for autowire)
* If there are more, you can specify one (for autowire)
* I'll add support for Target Attribute & named parameter

lyrixx:
Here we go! New version published. I added everything I wanted. I wrote some documentation. Can I have a review please? And if it's okay, I'll write some tests eventually

CHANGELOG.md
```diff
@@ -7,9 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add a bundle to ease integration with Symfony
```
Korbeil, line 10:
```suggestion
### Added
- Add a bundle to ease integration with Symfony
```

src/Bridge/Symfony/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,109 @@
...
+class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder('elastically');
+        $rootNode = $treeBuilder->getRootNode();
+
+        $rootNode
+            ->children()
+                ->arrayNode('connections')
```
Korbeil, line 27: Is it possible to get a minimum of 1 connection ? I'm not sure it's checked atm

Korbeil (approved):
Thanks a lot for the bundle, it looks amazing !

src/Bridge/Symfony/DependencyInjection/Configuration.php
lyrixx: I don't see why we should force at least one connection. The bundle must work as expected with or without connection. ATM, the bundle only provide DI for connection, but later it might do other things (profiler, logger, command, etc). So we must not enforce this constraint.

README.md
```diff
@@ -227,57 +227,99 @@ _Default to `null`._
...
-Just declare the proper service in `services.yaml`:
+You'll need to add the bundle in `bundles.php`:
+
+```php
+// config/bundles.php
+return [
+    // ...
+    JoliCode\Elastically\Bridge\Symfony\ElasticallyBundle::class => ['all' => true],
+];
+```
```
damienalexandre, line 240: **💭 thought:** ‏Could it be automatic with Symfony Flex? 🤔 Maybe not in the context of a bridge?

README.md
```diff
@@ -227,57 +227,99 @@ _Default to `null`._
...
-            $indexClassMapping:
-                indexName: My\AwesomeDTO
+# config/packages/elastically.yaml
+elastically:
+    connections:
+        default:
+            client:
+                host:                '%env(ELASTICSEARCH_HOST)%'
+                # If you want to use the Symfony Http Client:
+                transport:           '@JoliCode\Elastically\Transport\HttpClientTransport'
```
damienalexandre, line 252: **💭 thought:** ‏I think that is should be the default in a Symfony context.

README.md
lyrixx: Yes! I plan to explain everything here, and then submit a PR to the recipes repo, and then update this repo 👍🏼

README.md
lyrixx: What if the http-client is not installed, or installed but not registered?
And this configuration is far from perfect IMHO, because it use the default HttpClient, and the default client can have some very special setting that might not work with ES. (header, base_uri, etc)

damienalexandre (approved):

damienalexandre:
Thanks!

PR #91, opened 2021-11-10
Fix index name mapper getting pure index name frome alias

kennywick:
* getPureIndexName now properly deals with prefixed aliases

* add Tests for getPureIndexName

* add function getMappedIndices to get a list of valid indices

### DIFF
--- a/src/IndexNameMapper.php
+++ b/src/IndexNameMapper.php
@@ -65,6 +65,17 @@ class IndexNameMapper
             return $matches[1];
         }
 
+        $prefixLength = strlen($this->prefix);
+
+        if ($this->prefix && substr($fullIndexName, 0, $prefixLength) === $this->prefix) {
+            return substr($fullIndexName, $prefixLength + 1);
+        }
+
         return $fullIndexName;
+    }
+
+    public function getMappedIndices(): array
+    {
+        return array_keys($this->indexClassMapping);
     }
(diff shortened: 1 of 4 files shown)


damienalexandre:
If I understand your change correctly, the idea is to support indices that don't have a date in their name? Is that because the indices are not created via Elastically?

kennywick:
> If I understand your change correctly, the idea is to support indices that don't have a date in their name? Is that because the indices are not created via Elastically?

In part yes (that's what we're doing in our integration tests), but also to support getting an index by its alias. Let's take an example of an index `todo` with prefix `foo`
```
$index = $client->getIndex('todo'); 

//This would be 'foo_todo'
$index->getName();

//Then we want to migrate the index
$indexBuilder->migrate($index);
```
So in the `IndexBuilder::migrate()` function 
```
        //This currently returns 'foo_todo' instead of 'todo'
        $pureIndexName = $this->indexNameMapper->getPureIndexName($currentIndex->getName());

        //This is now passed the prefixed Index name instead of the pure index name
        $newIndex = $this->createIndex($pureIndexName);
```
This then expects a mapping file that is named `foo_todo_mapping.yaml` instead of the existing `todo_mapping.yaml`

damienalexandre:
Thanks a lot @kennywick :+1:

PR #95, opened 2021-12-01
update php-cs-fixer to v3 + tweak the configuration + apply it to the codebase

lyrixx:


tests/Jane/JaneTest.php
```diff
@@ -40,7 +47,8 @@ public function testCreateIndexAndSearchWithJaneObject()
         ];
         $encoders = [new JsonEncoder(
             new JsonEncode(),
-            new JsonDecode([JsonDecode::ASSOCIATIVE => true])),
+            new JsonDecode([JsonDecode::ASSOCIATIVE => true])
```
welcoMattic, line 50: Bizarre ça, non ?

tests/Transport/HttpClientTransportTest.php
```diff
@@ -58,21 +65,22 @@ public function testCreateIndex(): void
...
+                      "hits" : {
+                        "total" : {
+                          "value" : 0,
+                          "relation" : "eq"
+                        },
+                        "max_score" : null,
+                        "hits" : [ ]
+                      }
+                    }
+                    JSON
```
welcoMattic, line 82: IIRC, on Windows it could break the test. Do we need to care about?

lyrixx: I don't understand :/ why would it breaks test?

lyrixx:
Failure are not related. There are all deprecation notice.

lyrixx:
There are too much deprecation + type error. This lib does not seem compatibly with something.

PR #105, opened 2022-05-02
Fix Indexer to works with Models

### LINKED ISSUE #70: Elastica\Document constructor expects array|string, object given
Hi!

According to the docs, to add a document to the index, you have to to the following:
```php
// DTO is a PHP object
$dto = new \StdClass();

// Add a document to the queue
$indexer->scheduleIndex('beers', new Document('123', $dto));
$indexer->flush();
```

In the [indexer](https://github.com/jolicode/elastically/blob/master/src/Indexer.php#L39), if the `$data` inside the `Document` (the `$dto` parameter) is an object, it is sent to the `serializer`:
```php
if (\is_object($document->getData())) {
    $context = $this->client->getSerializerContext(\get_class($document->getData()));
    $document->setData($this->serializer->serialize($document->getData(), 'json', $context));
}
```

However, the [constructor](https://github.com/ruflin/Elastica/blob/master/src/Document.php#L43) of `Elastica\Document` is typehinted as `array|string`:
```php
/**
 * Creates a new document.
 *
 * @param string|null  $id    The document ID, if null it will be created
 * @param array|string $data  Data array
 * @param Index|string $index Index name
 */
public function __construct(?string $id = null, $data = [], $index = '')
```

And [`getData`](https://github.com/ruflin/Elastica/blob/master/src/Document.php#L203) returns an `array|string` as well:
```php
    /**
     * Returns the document data.
     *
     * @return array|string Document data
     */
    public function getData()
```

At the moment, this results in statistical analysis errors. In the future, this could lead to PHP errors when these PHPDoc's are converted to real typed properties.

laryjulien:
This PR fixes #70.

Far from perfect but it works. An other way of fixing this issue is to create an Elastically\Document that extends Elastica\Document and adding a property for the "model". I am not sure this will be a better solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ $dto->bar = 'American Pale Ale';
 $dto->foo = 'Hops from Alsace, France';
 
 // Add a document to the queue
-$indexer->scheduleIndex('beers', new Document('123', $dto));
+$indexer->scheduleIndex('beers', new Document('123', ['__model' => $dto]));
 $indexer->flush();
 
 // Set parameters on the Bulk
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -13,6 +13,7 @@ namespace JoliCode\Elastically;
 
 use Elastica\Bulk;
 use Elastica\Document;
+use Elastica\Exception\InvalidException;
 use Elastica\Index;
 use JoliCode\Elastically\Serializer\ContextBuilderInterface;
 use JoliCode\Elastically\Serializer\StaticContextBuilder;
@@ -42,10 +43,8 @@ class Indexer
     public function scheduleIndex($index, Document $document)
     {
         $document->setIndex($index instanceof Index ? $index->getName() : $index);
-        if (\is_object($document->getData())) {
-            $context = $this->contextBuilder->buildContext(\get_class($document->getData()));
-            $document->setData($this->serializer->serialize($document->getData(), 'json', $context));
-        }
+
+        $this->extractModel($document);
 
         $this->getCurrentBulk()->addDocument($document, Bulk\Action::OP_TYPE_INDEX);
 
@@ -64,10 +63,8 @@ class Indexer
     public function scheduleUpdate($index, Document $document)
     {
         $document->setIndex($index instanceof Index ? $index->getName() : $index);
-        if (\is_object($document->getData())) {
-            $context = $this->contextBuilder->buildContext(\get_class($document->getData()));
-            $document->setData($this->serializer->serialize($document->getData(), 'json', $context));
-        }
+
+        $this->extractModel($document);
 
         $this->getCurrentBulk()->addDocument($document, Bulk\Action::OP_TYPE_UPDATE);
 
@@ -77,10 +74,8 @@ class Indexer
     public function scheduleCreate($index, Document $document)
     {
         $document->setIndex($index instanceof Index ? $index->getName() : $index);
-        if (\is_object($document->getData())) {
-            $context = $this->contextBuilder->buildContext(\get_class($document->getData()));
-            $document->setData($this->serializer->serialize($document->getData(), 'json', $context));
-        }
+
+        $this->extractModel($document);
 
         $this->getCurrentBulk()->addDocument($document, Bulk\Action::OP_TYPE_CREATE);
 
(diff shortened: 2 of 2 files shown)


lyrixx:
I don't understand the move. With this PR, `Document::getData()` still return an object, but [the PHPDoc](https://github.com/ruflin/Elastica/blob/master/src/Document.php#L201) keeps saying it return an array or a string

What did I miss?

laryjulien:
I have created a specific key to store the object in Elastica\Document $data array. It must be use like this: `new Document($id, ['__model' => $object])`. 
Then I have changed Elastically\Indexer to look for __model property in Document. If all conditions are ok, it normalizes the object in __model and finally replaces the whole $data with normalized object. 
If the __model property is missing, then the normalize/replace will not be triggered.

lyrixx:
I got that part. But this does not solve the initial issue,isn't ?

laryjulien:
In a way, it does. 

$indexer->scheduleIndex('beers', new Document('123', $dto));
is replaced by 
$indexer->scheduleIndex('beers', new Document('123', ['_model' => $dto]));

So static analyzers are not yelling about the second parameter of Document being an object.

lyrixx:
OK, this solves only a part of the initial issue.

I'm sorry but I don't like your proposition. It really feels a hack, and it's not really natural.

I'll wait a bit for others feedback, but for now i'm voting -1 on this one.

laryjulien:
Does the solution with an Elastically\Document extending Elastica\Document by adding a property to store the object seems better to you?

lyrixx:
I guess, but this needs to be tested

laryjulien:
And to complete, my first solution is clearly a hack. Because you can not add an object to a Elastica\Document, $data is meant to be an array with Document properties and values.
I am only seeing two solutions: 
 - even we provide extra feature to Document by implementing a new one (my previous comment) 
 - or we add methods to Indexer that will take an object, normalize it, create a Document from it then doing the schedule part. 
 
 Others solutions are welcome!

lyrixx:
Maybe we can sent an PR upstream to fix the typehint

damienalexandre:
> Does the solution with an Elastically\Document extending Elastica\Document by adding a property to store the object seems better to you?

That mean we need to extend all the places where a Document is created by Elastica - we aim to override the least code possible.

> add methods to Indexer that will take an object, normalize it, create a Document from it then doing the schedule part

That could work for indexing, but that does not change the type mismatch in search results https://github.com/jolicode/elastically/blob/master/src/Result.php#L34.

From my POV it's quite acceptable to have an object in the "_data" property of Document even if it's supposed to be a `string|array` (the property is typed as array, the constructor as string + array by the way...): https://github.com/ruflin/Elastica/blob/master/src/Document.php#L20

laryjulien:
I can change type-hint on Elastica side but Elastica library will be more inconsistent due to $_data being an object. Example:
https://github.com/ruflin/Elastica/blob/master/src/Document.php#L263
And by traversing a lot of objects from Document, Index, Client, Bulk, AbstractDocument, IndexDocument to Action. This could be a problem here: https://github.com/ruflin/Elastica/blob/master/src/Bulk/Action.php#L45. 

If you have a use-case that justifies the type changing because this seems hard to argue about the benefits from Elastica POV

laryjulien:
https://github.com/ruflin/Elastica/issues/811
https://github.com/ruflin/Elastica/issues/1067

damienalexandre:
> If you have a use-case that justifies the type changing because this seems hard to argue about the benefits from Elastica POV

There is no advantage adding `object` for Elastica you are right.

But our use-case here is valid, Elastically first goal is to bring objects / DTO into Elastica. Sometimes it means ignoring the phpdoc. That may break in the future, for example if PHP Type hint are added in version 8 (discussion here https://github.com/ruflin/Elastica/discussions/1961#discussioncomment-1779120) but in the meantime it's very convenient to just create a new Document, pass the DTO and that's it. I would like to keep this DX as much as possible.

laryjulien:
What do you think about not type-hint $document in Indexer methods?
If $document is a Document, then it is added to the Bulk else we try to serialize it.

lyrixx:
IMHO this does not make any sense. We must not modify code for the sake of a tools. For now the code works as expected and because some PHP doc are wrong, some tools returns false positive. 

I think we are going in the wrong direction

lyrixx:
Hello, This has been solved in  #150 🎉

PR #106, opened 2022-05-04
Try using the Elastic Github Action + add ES 8

damienalexandre:
Inspired by https://github.com/elastic/elasticsearch-php/blob/master/.github/workflows/test.yml


lyrixx:
Nice! 💛

PR #107, opened 2022-06-22
FIX #56 Run the tests on a custom Elasticsearch port

### LINKED ISSUE #56: Change the port used in tests
We use the default 9200 port for all the tests, but as we run some destructive requests to Elasticsearch, maybe all tests should try to connect on another, safer port. 

That way there is no data destruction risk if someone runs the tests locally on his own ES server.

damienalexandre:
To avoid local destructions and other side effects.

See #56.


lyrixx (approved):

lyrixx:
Thanks

PR #118, opened 2022-10-07
Add OpenSearch in the test matrice

### LINKED ISSUE #113: Add OpenSearch in the test suite
As we are compatible with ES7 (and ES8 but not officially), we should also be compatible with OpenSearch.

This should be added to our test-suite too.

damienalexandre:
Fix #113.

Add OpenSearch 1 and 2 tests.

Had to duplicate the job because we can't have variables in the `uses` key.


lyrixx:
Super cool. Thanks! 💛

Issue #115, opened 2022-10-04
Remove deprecated code and Factory in favor of full DI


damienalexandre:
Dependency Injection should be the only way to deal with all our services.

For 2.0, let's get rid of all the error prone code and have only one way to use the lib - removing the Factory.

VincentLanglet:
Currently the factory allows to override multiple things with these options:
```
    public const CONFIG_BULK_SIZE = 'elastically_bulk_size';
    public const CONFIG_DENORMALIZER = 'elastically_denormalizer';
    public const CONFIG_INDEX_CLASS_MAPPING = 'elastically_index_class_mapping';
    public const CONFIG_INDEX_NAME_MAPPER = 'elastically_index_name_mapper';
    public const CONFIG_INDEX_PREFIX = 'elastically_index_prefix';
    public const CONFIG_MAPPINGS_DIRECTORY = 'elastically_mappings_directory';
    public const CONFIG_MAPPINGS_PROVIDER = 'elastically_mappings_provider';
    public const CONFIG_SERIALIZER = 'elastically_serializer';
    public const CONFIG_SERIALIZER_CONTEXT_BUILDER = 'elastically_serializer_context_builder';
    public const CONFIG_SERIALIZER_CONTEXT_PER_CLASS = 'elastically_serializer_context_per_class';
```

Am I wrong or some of them cannot be by the elastically configuration so far ?
I only see config for:
```
bulk_size
index_class_mapping
prefix
mapping_directory
mapping_provider_service
serializer. context_builder_service
serializer. context_mapping
```

So we're missing:
```
denormalizer
index_name_mapper
serializer
```
https://github.com/jolicode/elastically/blob/master/src/Bridge/Symfony/DependencyInjection/Configuration.php#L69

damienalexandre:
I think you are right :+1:

lyrixx:
I don't think we need an **open** issue for such things.

Instead, let's do like Symfony: We keep an UPGRADE.md file up to date to track all deprecation. Then when the 2.x branch is created, we drop everything.

lyrixx:
@VincentLanglet : We don't need bundle configuration for overriding a service. I'm a bit against adding options to override any part of the lib.

For exemple, the `bulk_size`, or the `mapping_directly` are values depending on the application structure / design.

The `denormalizer`, `serializer`, `index_name_mapper` are usually not changed. So we don't need an entry in the configuration to change tem.

---

[...]

PR #120, opened 2022-10-08
Ability to change mapping filename

insekticid:
if using non standardized mapping file naming

```php
//this will work
 $indexBuilder->migrate($this->elasticallyClient->getIndex($indexName), [], ['filename' => 'content_mapping.yaml']);

//this works
$index = $indexBuilder->createIndex($indexName, ['filename' => 'content_mapping.yaml']);
```

### DIFF
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -70,10 +70,10 @@ class IndexBuilder
         $index->getSettings()->setRefreshInterval('1s');
     }
 
-    public function migrate(Index $currentIndex, array $params = []): Index
+    public function migrate(Index $currentIndex, array $params = [], array $context = []): Index
     {
         $pureIndexName = $this->indexNameMapper->getPureIndexName($currentIndex->getName());
-        $newIndex = $this->createIndex($pureIndexName);
+        $newIndex = $this->createIndex($pureIndexName, $context);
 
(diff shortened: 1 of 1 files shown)


damienalexandre:
Thanks a lot - that was a bug indeed.

In the next major version I intend to remove this context (https://github.com/jolicode/elastically/issues/115) because it's a source of confusion and error - mapping providers are going to be more robust that way.

I will release this patch very soon. Thanks!

insekticid:
@damienalexandre Ok, so think about solution, how to use same mapping file for more indexes. I have indexes with prefix `content-` + langName (e.g. cs, de, sk...) so I do not want to make copies of content_mapping.yaml file with the same config. Or think about extending from base mapping file and use something like

```
content_cs_mapping.yaml:

imports:
  - { resource: ' content_mapping.yaml' }


content_de_mapping.yaml:

imports:
  - { resource: ' content_mapping.yaml' }
```

damienalexandre:
Yes, thanks for the suggestion :+1: 

We will do our best to include this in the YamlMappingProvider, and if not possible, you always have the possibility of building your own mapping provider with the https://github.com/jolicode/elastically/blob/master/src/Mapping/MappingProviderInterface.php.

PR #125, opened 2022-11-29
Throw exception if empty index in ResultSetBuilder #116

### LINKED ISSUE #116: ResultSetBuilder::buildModel(): Argument #2 ($indexName) must be of type string
ResultSetBuilder::buildModel() should accept array of indexes, when searching between multiple indexes

```
use JoliCode\Elastically\Client;

class ContentRepository
{
    public function __construct(
        private readonly Client $client)
    {
    }

   function getSearch(Query $query): Search  {
      $search = $this->client->getIndex('content-cs')->createSearch();
      $search->addIndexByName('content-de');
      return $search->setQuery($query);
  }
}

(new ContentRepository())->getSearch($query)->search();

```

[ResultSetBuilder::buildModel] Argument #2 ($indexName) must be of type string

Found problem here
https://github.com/ruflin/Elastica/blob/6d51cdbff2fb37aceac8b70adf9042ccb9be0aed/src/Result.php#L126

getParam() returns empty array when key not exists and yes, it is missing
https://github.com/ruflin/Elastica/blob/6d51cdbff2fb37aceac8b70adf9042ccb9be0aed/src/Result.php#L68
image

MrYamous:
Fix #116 
Using @insekticid suggestion for fix

### DIFF
--- a/src/ResultSetBuilder.php
+++ b/src/ResultSetBuilder.php
@@ -12,6 +12,7 @@
 namespace JoliCode\Elastically;
 
 use Elastica\Document;
+use Elastica\Exception\RuntimeException;
 use Elastica\Query;
 use Elastica\Response;
 use Elastica\ResultSet;
@@ -67,6 +68,10 @@ class ResultSetBuilder implements BuilderInterface
 
     private function buildModelFromResult(Result $result)
     {
+        if (!$result->getIndex()) {
+            throw new RuntimeException('Returned index is empty. Check your "filter_path".');
+        }
+
         return $this->buildModel($result->getSource(), $result->getIndex(), [
             self::RESULT_KEY => $result,
         ]);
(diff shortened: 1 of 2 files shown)


lyrixx:
Can you rebase to fix the conflict? Thanks

src/ResultSetBuilder.php
```diff
@@ -67,6 +68,10 @@ public function buildModelFromDocument(Document $document)
 
     private function buildModelFromResult(Result $result)
     {
+        if (!$result->getIndex()) {
+            throw new RuntimeException('Returned index is empty. Check your filter_path');
```
lyrixx, line 72:
```suggestion
            throw new RuntimeException('Returned index is empty. Check your "filter_path".');
```

MrYamous:
Should be better now @lyrixx

PR #138, opened 2023-01-13
CHANGELOG: remove extra "support" word

jmsche:


lyrixx:
good catch, thanks 👍🏼

PR #144, opened 2023-01-16
Fix bundle issues

### LINKED ISSUE #126: Symfony 6.2 deprecation
Hi,

Since I upgraded a project to Symfony 6.2 I have this deprecation:

```
Since symfony/messenger 6.2: Implementing "Symfony\Component\Messenger\Handler\MessageHandlerInterface" is deprecated, use the "Symfony\Component\Messenger\Attribute\AsMessageHandler" attribute instead. 
```

From this class: https://github.com/jolicode/elastically/blob/f918cd0465508d456cf4314a9ce0e278efc3ff2b/src/Messenger/IndexationRequestHandler.php

### LINKED ISSUE #143: IndexationRequestHandler issues
Hi,

When setting up latest (dev-master) code of Elastically on a project to test Symfony bridge, I get the following error when using Messenger:

```
Cannot autowire service "JoliCode\Elastically\Messenger\IndexationRequestHandler": argument "$indexNameMapper" of method "__construct()" references class "JoliCode\Elastically\IndexNameMapper" but no such service exists. You should maybe alias this class to the existing "elastically.default.index_name_mapper" service.
```

Setting the value of `$indexNameMapper` to the one provided by the error fixes it, but I'm wondering if it should not be autowired by the bridge instead?

Else docs should be updated.

Korbeil:
This PR will do:
- Fixes #126 by removing the interface and using the new attribute on the Handler
- Fixes #143 by adding the default IndexNameMapper in the Symfony DI in the Bundle

### DIFF
--- a/src/Bridge/Symfony/DependencyInjection/ElasticallyExtension.php
+++ b/src/Bridge/Symfony/DependencyInjection/ElasticallyExtension.php
@@ -14,6 +14,7 @@ namespace JoliCode\Elastically\Bridge\Symfony\DependencyInjection;
 use JoliCode\Elastically\Client;
 use JoliCode\Elastically\IndexBuilder;
 use JoliCode\Elastically\Indexer;
+use JoliCode\Elastically\IndexNameMapper;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ChildDefinition;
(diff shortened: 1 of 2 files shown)


damienalexandre:
Thanks a lot!

jmsche:
Thanks @Korbeil, everything seems fine now :)

PR #149, opened 2023-02-08
Get ride of bad typing in Document::getData()

### LINKED ISSUE #148: Document->setData() is not supposed to accept DTO.
Hi @damienalexandre, @lyrixx

In the doc, there is the following example
```
$dto = new Beer();
$dto->bar = 'American Pale Ale';
$dto->foo = 'Hops from Alsace, France';

// Add a document to the queue
$indexer->scheduleIndex('beers', new Document('123', $dto));
$indexer->flush();
```
but this is throwing static analysis issues because the second parameter `$data` is supposed to be a string|array
https://github.com/ruflin/Elastica/blob/8.x/src/Document.php#L43

This also means that one day, when typehint will be added to elastica, the solution with DTO won't work anymore.

I was surprised it currently works, but discovered it was because of the following code
https://github.com/jolicode/elastically/blob/master/src/Indexer.php#L45-L48, but this won't work the day native typehint will be added to `setData/getData` methods.

I'm not familiar with the library and the strategy behind but a proper way to change this could be:
- Exposing a method `prepareData` which do
```
$context = $this->contextBuilder->buildContext(\get_class($object));
return $this->serializer->serialize($object, 'json', $context);
```
- Updating the doc to do
```
$indexer->scheduleIndex('beers', new Document('123', $indexer->prepareData($dto)));
```
- Deprecating the `\is_object` check in `scheduleIndex`.

Another way could be to change the `scheduleIndex` signature from 
```
scheduleIndex($index, Document $document)
```
to 
```
public function scheduleIndex(string|Index $index, ?string $id, string|data|object $data))
    {
        if (\is_object($data)) {
            $context = $this->contextBuilder->buildContext(\get_class($data));
            $data = $this->serializer->serialize($data, 'json', $context);
        }

		$document = new Document($id, $data, $index);
        $this->getCurrentBulk()->addDocument($document, Bulk\Action::OP_TYPE_INDEX);

        $this->flushIfNeeded();
    }
```

WDYT ? Do you see another way solution to
- Avoid static analysis issues
- Avoid this library to loose the "data as DTO" feature when elastica add `array|string` as native type.

lyrixx:
fixes #148

### DIFF
--- a/src/Indexer.php
+++ b/src/Indexer.php
@@ -12,8 +12,9 @@
 namespace JoliCode\Elastically;
 
 use Elastica\Bulk;
-use Elastica\Document;
+use Elastica\Document as ElasticaDocument;
 use Elastica\Index;
+use JoliCode\Elastically\Model\Document;
 use JoliCode\Elastically\Serializer\ContextBuilderInterface;
 use JoliCode\Elastically\Serializer\StaticContextBuilder;
 use Symfony\Component\Serializer\SerializerInterface;
@@ -39,13 +40,10 @@ class Indexer
         $this->contextBuilder = $contextBuilder ?? new StaticContextBuilder();
     }
 
-    public function scheduleIndex($index, Document $document)
+    public function scheduleIndex($index, Document|ElasticaDocument $document)
     {
         $document->setIndex($index instanceof Index ? $index->getName() : $index);
-        if (\is_object($document->getData())) {
-            $context = $this->contextBuilder->buildContext(\get_class($document->getData()));
-            $document->setData($this->serializer->serialize($document->getData(), 'json', $context));
-        }
+        $this->updateDocumentData($document);
 
         $this->getCurrentBulk()->addDocument($document, Bulk\Action::OP_TYPE_INDEX);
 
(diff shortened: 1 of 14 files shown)


src/Indexer.php
```diff
@@ -42,10 +41,7 @@ public function __construct(Client $client, SerializerInterface $serializer, int
     public function scheduleIndex($index, Document $document)
```
VincentLanglet: You might need to support 
```
public function scheduleIndex($index, Document|Elastica\Document $document)
```
in 1.x versions in order to be BC.

src/Indexer.php
```diff
@@ -169,4 +161,17 @@ private function refreshBulkRequestParams()
...
     }
+
+    private function updateDocumentData($document): void
+    {
+        if ($document instanceof Document) {
+            $context = $this->contextBuilder->buildContext(\get_class($document->getDto()));
+            $data = $this->serializer->serialize($document->getData(), 'json', $context);
+            $document->setData($data);
+        } elseif ($document instanceof ElasticaDocument) {
+            // @phpstan-ignore-next-line
```
VincentLanglet, line 172:
```suggestion
        } elseif ($document instanceof ElasticaDocument && \is_object($document->getData())) {
```

To keep support of string|array

src/Indexer.php
```diff
@@ -169,4 +161,17 @@ private function refreshBulkRequestParams()
             $this->currentBulk->setRequestParam($key, $value);
         }
     }
+
+    private function updateDocumentData($document): void
+    {
+        if ($document instanceof Document) {
+            $context = $this->contextBuilder->buildContext(\get_class($document->getDto()));
+            $data = $this->serializer->serialize($document->getData(), 'json', $context);
```
VincentLanglet, line 169:
```suggestion
            $data = $this->serializer->serialize($document->getDto(), 'json', $context);
```

PR #150, opened 2023-02-08
Get ride of bad typing in Document::getData()

VincentLanglet:
This is the PR https://github.com/jolicode/elastically/pull/149 with green CI.


src/ResultSetBuilder.php
```diff
@@ -60,14 +59,9 @@ public function buildModelFromIndexAndData(string $indexName, $source)
         return $this->buildModel($indexName, $source, []);
     }
 
-    public function buildModelFromDocument(Document|ElasticaDocument $document)
+    public function buildModelFromDocument(Document $document)
```
lyrixx, line 62: We must keep that to preserve the BC

lyrixx (commented):
Thank you very much for taking care of this PR!

I lake some time to finish mine :D 

your last commit is cool, but we really need to preserve the BC, we don't want to bump to a major version if we can avoid id!

Can you ensure it? 

Thanks

VincentLanglet:
> Can you ensure it?

Sure

src/ResultSetBuilder.php
VincentLanglet: I renamed the imported class. This is BC since I'm supporting `ElasticaDocument`.

src/Model/Document.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace JoliCode\Elastically\Model;
+
+use Elastica\Document as ElasticaDocument;
+
+class Document extends ElasticaDocument
+{
+    private ?object $dto = null;
+
+    public function __construct($id, object $dto = null, $data = [], $index = '')
```
lyrixx, line 20: WDYT about enforcing the type on data and index? (put the same as in the PHPDoc of ElasticaDocument?)

tests/IndexerTest.php
```diff
@@ -183,6 +184,7 @@ public function testIndexJsonString(): void
 
         $indexer->scheduleIndex($indexName, new Document(
```
lyrixx, line 185: We need a test with old code path, and the new one

lyrixx (commented):
What do you think about always returning an Elastically\Document from the index too?

src/Model/Document.php
VincentLanglet: Sure

VincentLanglet:
> What do you think about always returning an Elastically\Document from the index too?

Sure, I thought about this too.
If you look at the last commit, this is what I tried to do.
But if you run the tests, you'll see that I get two errors, like
```
Index Builder (JoliCode\Elastically\Tests\IndexBuilder)
 ✘ Migrate
   ┐
   ├ Elastica\Exception\RuntimeException: Unknown class for index "empty". Please check your configuration.
   │
   ╵ /Users/vincentl/Perso/elastically/src/IndexNameMapper.php:50
   ╵ /Users/vincentl/Perso/elastically/src/ResultSetBuilder.php:86
   ╵ /Users/vincentl/Perso/elastically/src/ResultSetBuilder.php:64
   ╵ /Users/vincentl/Perso/elastically/src/Index.php:34
   ╵ /Users/vincentl/Perso/elastically/tests/IndexBuilderTest.php:240
```

Any idea why ?

lyrixx:
Arf, no :/

VincentLanglet:
I reverted the last commit since returning an `Document` instead of `ElasticaDocument` is not straight forward.
It can be done in another PR if wanted, this will still be BC.

src/ResultSetBuilder.php
```diff
@@ -59,7 +59,7 @@ public function buildModelFromIndexAndData(string $indexName, $source)
         return $this->buildModel($indexName, $source, []);
     }
 
-    public function buildModelFromDocument(Document $document)
+    public function buildModelFromDocument(ElasticaDocument $document)
```
lyrixx, line 62: This is a BC break. We should revert this line, and deal with both class. But I guess there is nothing to do apart reverting this line

lyrixx (commented):
This is almost good for me 👍🏼 ; There is only one little comment.

Can you squash your commits when everything is OK? Thanks

src/ResultSetBuilder.php
damienalexandre: No it's not :thinking: it's the same signature, only the imported named changed :thinking:

lyrixx: Oups 🤦🏼 Thanks

damienalexandre:
Hello! Thanks a lot for this contribution!

**💭 thought:** I'm wondering something about DX. If I'm working with `JoliCode\Elastically\Model\Document` when indexing, I'm expecting the same class when calling `$results = $client->getIndex('beers')->getDocument('123');`. I will get a Document but calling `getDto` on it will explode (because we get a `Elastica\Document` I think). This is bad for consistency / reliability and it's error prone for users of this library, don't you think?

Also, we have a `getModel();` method:

https://github.com/jolicode/elastically/blob/5de5a3131928e13c32f6fb3f182749a3930b6a3f/src/Result.php#L21-L24

Would be nice to use the same terminology here (dto = model).

lyrixx:
Can you also update the `CHANGELOG.md` ?

VincentLanglet:
> Hello! Thanks a lot for this contribution!
> 
> **💭 thought:** I'm wondering something about DX. If I'm working with `JoliCode\Elastically\Model\Document` when indexing, I'm expecting the same class when calling `$results = $client->getIndex('beers')->getDocument('123');`. I will get a Document but calling `getDto` on it will explode (because we get a `Elastica\Document` I think). This is bad for consistency / reliability and it's error prone for users of this library, don't you think?

Yes I agree.
But I got the following issue (https://github.com/jolicode/elastically/pull/150#issuecomment-1422891812) when trying to do this change.

The fact is that, currently
```
$client->getIndex('beers')->getDocument('123')
```
doesn't require to add a mapping config, but
```
$client->getIndex('beers')->getModel('123')
```
does.

Doing the change (returning `JoliCode\Elastically\Model\Document`) will imply that
```
$client->getIndex('beers')->getDocument('123')
```
will require the mapping config too ; and this is a BC break (but could done for 2.x).

> Also, we have a `getModel();` method:
> 
> https://github.com/jolicode/elastically/blob/5de5a3131928e13c32f6fb3f182749a3930b6a3f/src/Result.php#L21-L24
> 
> Would be nice to use the same terminology here (dto = model).

Do you prefer me to add a new `getDto()` or to change the `JoliCode\Elastically\Model\Document` to rename the `dto` property to `model` ?

damienalexandre:
Yes having `private ?object $dto;` renamed to `model` would be preferable (like this we don't introduce a new name).

Thanks!

VincentLanglet:
> Let's goooooo 🎉
> 
> Thanks a lot for this improvement!

Thanks. Do you have time to create a release with it @damienalexandre ?

damienalexandre:
We would like to sit on it / wait for feedback / test it on real project first :pray:  if that's ok for you?

VincentLanglet:
> We would like to sit on it / wait for feedback / test it on real project first 🙏 if that's ok for you?

sure

VincentLanglet:
Did you have time to try it on real project @lyrixx, @damienalexandre ?
This seems to be fine on my side

PR #152, opened 2023-02-17
doc(symfony): Provide a working example out of the box

damienalexandre:
The example in the documentation was not working out of the box.

The HttpClientTransport MUST be declared first:

```yaml
JoliCode\Elastically\Transport\HttpClientTransport: ~
```

So I think it's better to comment it out.

Also, what about forcing this transport when using Symfony?!

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,8 +248,8 @@ elastically:
         default:
             client:
                 host:                '%env(ELASTICSEARCH_HOST)%'
-                # If you want to use the Symfony Http Client:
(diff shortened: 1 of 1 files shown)


lyrixx (approved):

Korbeil (approved):

lyrixx:
> Also, what about forcing this transport when using Symfony?!

nope :) you probabily one specific http client, not the default one

lyrixx:
Thanks

PR #155, opened 2023-03-08
feature(IndexBuilder): add dry mode in function purgeOldIndices

vgreb:
Hi there,

In a project, I wanted to remove old indices on my ES cluster. However I did not want to run `IndexBuilder::purgeOldIndices()` in live mode risking to wrongly delete an indice.

Here's why a propose to add a dry mode on `IndexBuilder::purgeOldIndices()` to allow listing which operations would be performed if ran in live mode.

Do you think this would be usefull / relevant ?

### DIFF
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -96,7 +96,7 @@ class IndexBuilder
         return $newIndex;
     }
 
-    public function purgeOldIndices(string $indexName): array
+    public function purgeOldIndices(string $indexName, bool $dryRun = false): array
     {
         $indexName = $this->indexNameMapper->getPrefixedIndex($indexName);
 
@@ -149,13 +149,17 @@ class IndexBuilder
 
             if ($livePassed && $afterLiveCounter > 1) {
                 // Remove
-                $index = new \Elastica\Index($this->client, $realIndexName);
-                $index->delete();
+                if (false === $dryRun) {
+                    $index = new \Elastica\Index($this->client, $realIndexName);
+                    $index->delete();
+                }
                 $operations[] = sprintf('%s deleted.', $realIndexName);
             } elseif ($livePassed && 1 === $afterLiveCounter) {
                 // Close
-                $index = new \Elastica\Index($this->client, $realIndexName);
-                $index->close();
+                if (false === $dryRun) {
+                    $index = new \Elastica\Index($this->client, $realIndexName);
+                    $index->close();
+                }
                 $operations[] = sprintf('%s closed.', $realIndexName);
             }
         }
--- a/tests/IndexBuilderTest.php
+++ b/tests/IndexBuilderTest.php
@@ -206,6 +206,38 @@ final class IndexBuilderTest extends BaseTestCase
         $this->assertCount(0, $operations);
     }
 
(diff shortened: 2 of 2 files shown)


src/IndexBuilder.php
```diff
@@ -96,7 +96,7 @@ public function migrate(Index $currentIndex, array $params = [], array $context
         return $newIndex;
     }
 
-    public function purgeOldIndices(string $indexName): array
+    public function purgeOldIndices(string $indexName, bool $dry = false): array
```
lyrixx, line 99: Can you rename this parameter to `dryRun`, it's more standard

tests/IndexBuilderTest.php
```diff
@@ -206,6 +206,38 @@ public function testPurgerDistinguishesIndicesWithTheSamePrefix(): void
         $this->assertCount(0, $operations);
     }
 
+    /** @dataProvider purgeIndexBuilderProvider */
+    public function testDoNotPurgeAndCloseOldIndicesInDryMode(IndexBuilder $indexBuilder): void
+    {
+        $index1 = $indexBuilder->createIndex('hop');
+        $this->assertInstanceOf(Index::class, $index1);
+
+        usleep(1200000); // 1,2 second
```
lyrixx, line 215: I wonder why you need to sleep here?
If you need it because we use the current timestamps in the index name, you can use the symfony helper for that https://symfony.com/doc/current/components/phpunit_bridge.html#time-sensitive-tests

It allows to run test as fast as possible

lyrixx (commented):
Thanks for your first contribution 🎉

I let two minor comment, otherwise it's good. Thanks again

tests/IndexBuilderTest.php
vgreb: To be honest, I don't know why there is a need for a sleep here.
My test is largely, even entirely, inspired by @damienalexandre 's [testPurgeAndCloseOldIndices](https://github.com/jolicode/elastically/blob/master/tests/IndexBuilderTest.php#L144-L181) test.

However, I will update my test in order to use the symfony helper.

src/IndexBuilder.php
vgreb: I've renamed the parameter to `dryRun`.

tests/IndexBuilderTest.php
vgreb: I've added the `@group time-sensitive` annotation on my test.

lyrixx: Wonderful :)

lyrixx (approved):

damienalexandre:
Sorry, you need to remove the `time-sensitive` annotation to fix the CI:

> Elastica\Exception\RuntimeException: Index "hop_2023-03-09-151840" is already created, something is wrong.

vgreb:
> Sorry, you need to remove the `time-sensitive` annotation to fix the CI:

No problem, it's done !

lyrixx:
I'll fix broken test in another PR (not your fault)

damienalexandre (approved):

damienalexandre:
Thanks a lot for this nice addition :+1:

PR #159, opened 2023-03-28
Fix buildModel argument order when called via buildModelFromIndexAndData

benji07:
I was trying to extends your ResultSetBuilder to include data from the suggestion and i saw that the buildModel argument where in an incorrect order


lyrixx:
Oh! That's strange. Can you add a test please?

benji07:
@lyrixx done

damienalexandre (approved):
Thanks for this correction :+1: 

I just discovered this method, and it's not used in the library itself. Are you using it yourself? (you said to hydrate suggestion results?).

benji07:
I added a custom ResultSetBuilder to handler Suggestion, but they are not return in the hits entry, bug in the suggest entry.

```
$suggest = new Suggest();
$suggest->addSuggestion(
    (new Completion('nameSuggester', 'name.completion'))
        ->setPrefix($search)
        ->setSize($this->query->getParam('size'))
        ->setFuzzy(['fuzziness' => 1]),
);
$this->query->setSuggest($suggest);
```

```
class ResultSetBuilder extends BaseResultSetBuilder
{
    public function buildResultSet(Response $response, Query $query): ResultSet
    {
        return new ResultSet($response, $query, $this->buildResult($response));
    }

    private function buildResult(Response $response): array
    {
        $data = $response->getData();

        $results = [];

        if (isset($data['hits']['hits'])) {
            foreach ($data['hits']['hits'] as $hit) {
                $resultModel = new Result($hit);
                $resultModel->setModel($this->buildModelFromIndexAndData((string) $resultModel->getIndex(), $resultModel->getSource()));

                $results[$resultModel->getModel()->slug] = $resultModel;
            }
        }

        if (isset($data['suggest']['nameSuggester'][0]['options'])) {
            foreach ($data['suggest']['nameSuggester'][0]['options'] as $hit) {
                $resultModel = new Result($hit);
                $resultModel->setModel($this->buildModelFromIndexAndData((string) $resultModel->getIndex(), $resultModel->getSource()));

                if (!isset($results[$resultModel->getModel()->slug])) {
                    $results[$resultModel->getModel()->slug] = $resultModel;
                }
            }
        }

        return array_values($results);
    }
}
```

damienalexandre:
Thanks a lot again :+1: nice use-case too!

PR #160, opened 2023-04-12
Add missing @throws annotation

VincentLanglet:
Hi @lyrixx @damienalexandre,

I recently added a lot of missing `@throws` tag to the Elastica lib with https://github.com/ruflin/Elastica/pull/2153
I saw the same were missing in Elastically so here is the PR.

When using PHPStan, it can report exceptions which are not caught if and only if the library is correctly using the @throws annotations. Same for PHPStorm. So this PHPDoc is really helpful.

### DIFF
--- a/src/IndexNameMapper.php
+++ b/src/IndexNameMapper.php
@@ -11,6 +11,7 @@
 
 namespace JoliCode\Elastically;
 
+use Elastica\Exception\ExceptionInterface;
 use Elastica\Exception\RuntimeException;
 
 class IndexNameMapper
@@ -33,6 +34,9 @@ class IndexNameMapper
         return $name;
     }
 
+    /**
+     * @throws ExceptionInterface
+     */
     public function getIndexNameFromClass(string $className): string
     {
         $indexName = array_search($className, $this->indexClassMapping, true);
@@ -44,6 +48,9 @@ class IndexNameMapper
         return $this->getPrefixedIndex($indexName);
     }
 
+    /**
+     * @throws ExceptionInterface
+     */
     public function getClassFromIndexName(string $indexName): string
     {
         if (!isset($this->indexClassMapping[$indexName])) {
--- a/src/Messenger/IndexationRequestHandler.php
+++ b/src/Messenger/IndexationRequestHandler.php
@@ -12,6 +12,7 @@
 namespace JoliCode\Elastically\Messenger;
 
 use Elastica\Exception\Bulk\ResponseException;
+use Elastica\Exception\ExceptionInterface;
 use Elastica\Exception\RuntimeException;
 use JoliCode\Elastically\Client;
 use JoliCode\Elastically\Indexer;
@@ -54,6 +55,9 @@ class IndexationRequestHandler
         $this->client->setLogger(new NullLogger());
     }
 
+    /**
+     * @throws ExceptionInterface
+     */
     public function __invoke(IndexationRequestInterface $message)
     {
         $messages = [];
@@ -107,6 +111,9 @@ class IndexationRequestHandler
         }
     }
 
+    /**
+     * @throws ExceptionInterface
+     */
     private function schedule(Indexer $indexer, IndexationRequest $indexationRequest)
     {
         try {
--- a/src/Bridge/Symfony/Resources/config/services.php
+++ b/src/Bridge/Symfony/Resources/config/services.php
@@ -75,6 +75,5 @@ return static function (ContainerConfigurator $container) {
                 '$client' => abstract_arg('elastically.abstract.client'),
                 '$indexNameMapper' => abstract_arg('elastically.abstract.index_name_mapper'),
             ])
-
     ;
 };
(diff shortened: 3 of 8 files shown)


src/IndexNameMapper.php
```diff
@@ -44,6 +48,9 @@ public function getIndexNameFromClass(string $className): string
         return $this->getPrefixedIndex($indexName);
     }
 
+    /**
+     * @throws ExceptionInterface
+     */
     public function getClassFromIndexName(string $indexName): string
```
damienalexandre, line 54: **💭 thought:** ‏Does PHPStan **really** need this PHPDoc? 

I mean, there is no hidden code here, the method is clear about which Exception can occurs.

For example LEVEL 9 does not trigger any warning here: https://phpstan.org/r/6a1d0c05-47f0-42ec-bae9-acca046c4908

src/Messenger/IndexationRequestHandler.php
```diff
@@ -54,6 +55,9 @@ public function __construct(Client $client, MessageBusInterface $bus, DocumentEx
         $this->client->setLogger(new NullLogger());
     }
 
+    /**
+     * @throws ExceptionInterface
+     */
```
damienalexandre, line 60: **💭 thought:** ‏We only call `schedule()` here, that method already have a `@throw`, are you sure we need to put this annotation on the whole chain of calls!?

damienalexandre (commented):
Hi!

Thanks a lot for your contribution but I fail to see the issue you are resolving?

Is PHPStan not clever enough to read the chain of method calls and get the Exceptions from that?

This adds A LOT of phpdoc and it's adds a burden on the maintainer of this library if merged. I see some parts that are not covered also: must we add all the throws for Messenger Bus exceptions too? What about the HttpClient ones? 

Could you elaborate on the issue this will resolve? 

Thanks a lot!

src/IndexNameMapper.php
VincentLanglet: The playground is not a good proof, because the exception analysis is not enabled on it.

PHPStan is only looking at the code inside the function when he's doing the analysis of this function. But when analysis is done, he's only looking at phpdoc like the method was coming from an interface. Especially because someone could extend the class and override the method implementation.

In the same way, no error is reported on the `test` method for https://phpstan.org/r/2fbe6b49-15e6-473a-b9f3-aceb4676abdd.

So yes, PHPStan really need the PHPDoc (and in a general way, always need the phpdoc on interface or methods)

VincentLanglet:
> Hi!
> 
> Thanks a lot for your contribution but I fail to see the issue you are resolving?
> 
> Is PHPStan not clever enough to read the chain of method calls and get the Exceptions from that?
> 
> This adds A LOT of phpdoc and it's adds a burden on the maintainer of this library if merged. I see some parts that are not covered also: must we add all the throws for Messenger Bus exceptions too? What about the HttpClient ones?
> 
> Could you elaborate on the issue this will resolve?
> 
> Thanks a lot!

I tried to be clear with https://github.com/jolicode/elastically/pull/160#discussion_r1164483821, tell me if you have any question.

For the HttpClient exceptions, you mean https://github.com/jolicode/elastically/blob/master/src/Transport/HttpClientTransport.php#L51 ?
The annotation will be on the original method in the AbstractTransport.

For the MessengerBus, I'm not sure of which class we talk about.

For this PR, I was focusing on all the Elastica-related exceptions.

VincentLanglet:
> * if we add `@throws`, I think we must **add them ALL** - not only Elastica's - otherwise the information we communicate to PHPStan is wrong! For example `getModel($id)` can throw all of this too:

I can add more annotation if necessary/wanted. I added the most useful.

>  Like I said there is also some HttpClient call. We use Messenger. We use the Yaml parser. We use the Serializer... All of those "external code" usages can and will throw Exceptions. And they have their own release cycle, their own life: they may change Exceptions.

That's why I used the interface for ElasticaException, it's allow more flexibility.
Moreover, I already updated a lot of codebase with `@throws` annotation, and it never changed/changed a lot.
Keep in mind that throwing a new Exception would be a BC-break, so it only happen for new majors versions.

> I think it's better to have no information that having a partial one and missing some important runtime issues.
>
> I talk about it internally too, and we agreed this is not a path we want to pursue for this library.
> 
> I'm sorry to let you down on this change - I hate to reject open-source contributions - but as we maintain this library on the long run we don't want to add those annotations if they are not 100% bulletproof.

I'm really sad about this decision, and would recommend strongly to reconsider this decision. 
No `@throws` annotation is way worst than "almost all the throws annotation".
To me it's the same than saying than it's better to not use any `@param/@return` annotation because they might change.

Let me give you a simple example. Because of those missing annotation in elastically code we missed to use a `try/catch` somewhere in our code ; this result in a recent crash in prod with data lost. All of this could have be avoided with those annotation because PHPStan would have caught up the missing try/catch.
Without this PHPDoc we cannot safely use elastically code without the fear of missing a try/catch later...

damienalexandre:
We think this could be a good way to deal with this yes! Didn't know it was possible 👍 

Thanks for your perseverance :-)

PR #165, opened 2023-05-16
Do not wait for Elasticsearch reindex to finish

Korbeil:
I have a use case where I don't care that the Elasticsearch reindex is done before switching the index as live. So I would like to make this change.

Hope this is fine to you guys ? :)

### DIFF
--- a/src/IndexBuilder.php
+++ b/src/IndexBuilder.php
@@ -87,7 +87,7 @@ class IndexBuilder
         if (!$waitForCompletion) {
             return $newIndex;
         }
-        
+
         $taskId = $response->getData()['task'];
 
         $task = new Task($this->client, $taskId);


src/IndexBuilder.php
```diff
@@ -84,6 +84,10 @@ public function migrate(Index $currentIndex, array $params = [], array $context
             throw new RuntimeException(sprintf('Reindex call failed. %s', $response->getError()));
         }
 
+        if (!$waitForCompletion) {
```
renanbr, line 87: You might also call `$reindex->setWaitForCompletion()` as well.

This may push some additional change in this method. The `setWaitForCompletion()` could always receive `false`, then this method performs what it needs based on the `$waitForCompletion`. Or we also could call `setWaitForCompletion()` passing `$waitForCompletion` value and let the lib do the "wait" for us.

lyrixx:
Thanks for the PR! I have two concerns about the new early return in `migrate()` when `$waitForCompletion` is `false` (`src/IndexBuilder.php`):

1. **The reindex task ID is discarded.** The method returns before reading `$response->getData()['task']`, so the caller has no way to track progress, know when the reindex is done, or detect a failure. If the reindex fails or is still running, `markAsLive()` will switch the alias to an incomplete index without anyone noticing. It would be better to return the task ID (or an `Elastica\Task` object) to the caller.

2. **The old index can be deleted while the reindex is still reading from it.** With the usual flow `migrate(..., false)` → `markAsLive()` → `purgeOldIndices()`, once the alias has moved, `purgeOldIndices()` considers the source index as old and deletes it, while the server-side reindex is still copying from it. The new live index stays partial and the original data is lost. At minimum, the documentation should state that the task must be completed before calling `markAsLive()` / `purgeOldIndices()`, ideally with a way to wait for it using the task returned in point 1.

Also, the PR currently has conflicts with `master` and needs a rebase.

PR #169, opened 2023-05-26
:memo: Add a mini-doc about the prefix option

Nek-:
Better than nothing, right? 😇

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,9 @@ elastically:
                 # Fill a static context
                 context_mapping:
                     foo:                 bar
+
+            # If you want to add a prefix for your index in elasticsearch (you can still call it by its base name everywhere!)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -266,6 +266,9 @@ elastically:
                 # Fill a static context
                 context_mapping:
                     foo:                 bar
+
+            # If you want to add a prefix for your index in elasticsearch (you can still call it by it's base name everywhere!)
```
lyrixx, line 270:
```suggestion
            # If you want to add a prefix for your index in elasticsearch (you can still call it by its base name everywhere!)
```

lyrixx (approved):

damienalexandre:
Thanks a lot

PR #196, opened 2024-11-13
chore(upgrade): Prepare major version 2 by removing all the dead code

damienalexandre:
Following the merge of Elastica 8, we drop OpenSearch.

Also it's time to remove the deprecated features.

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -11,6 +11,10 @@
 
 namespace JoliCode\Elastically;
 
+use Elastic\Elasticsearch\Exception\ClientResponseException;
+use Elastic\Elasticsearch\Exception\MissingParameterException;
+use Elastic\Elasticsearch\Exception\ServerResponseException;
+use Elastic\Transport\Exception\NoNodeAvailableException;
 use Elastica\Exception\ExceptionInterface;
 use Elastica\Index as ElasticaIndex;
 use Elastica\ResultSet\BuilderInterface;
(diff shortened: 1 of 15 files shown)


src/Index.php
```diff
@@ -29,10 +33,16 @@ public function __construct(Client $client, string $name, ResultSetBuilder $resu
     }
 
     /**
+     * @return mixed|string|null
+     *
+     * @throws ClientResponseException
```
VincentLanglet, line 38: Instead listing all the exceptions, you might prefer to just add the interface
```
@throws ElasticsearchException
```
in the phpdoc. (Same for others phpdoc where you added `@throws`).

src/Index.php
```diff
@@ -29,10 +33,16 @@ public function __construct(Client $client, string $name, ResultSetBuilder $resu
     }
 
     /**
+     * @return mixed|string|null
```
VincentLanglet, line 36: This phpdoc seems useleess since there is `mixed` as native return type.

src/Index.php
damienalexandre: You have a point but when looking at the different exceptions, I see some value in declaring them independently.

We have 4 from elasticsearch/elasticsearch-php:

- ClientResponseException, which has `use ResponseTrait;` which gives hints on the HTTP Response
- ServerResponseException, which has `use ResponseTrait;` which gives hints on the HTTP Response
- MissingParameterException :white_check_mark: ok to use the ElasticsearchException
- NoNodeAvailableException :x: doesn't implement ElasticsearchException

And 2 from other deps:

- ExceptionInterface is from Elastica
- SerializerExceptionInterface is from the Serializer

I think it's nicer to keep them all, so static analysis can warn if someone tries a `$exception->getResponse()` on the wrong one, and still our auto-completion works when developing with it.

That's also used in Elastica itself:

https://github.com/ruflin/Elastica/blob/58042887616eeb63621412c03bc903056bbcee7e/src/Bulk.php#L299-L306

damienalexandre:
Thanks @VincentLanglet for the review :wink:

PR #204, opened 2024-12-13
Support for script fields

prugala:
### Support for script fields

Merge `_source` with `fields` before denormalization process.

Creating `script_fields` in search result is stored in `fields` property e.g.:

Search object:
```
{
"script_fields": {
        "foo": {
            "script": {
                "source": "params._source.bar == 'test'",
            }
        },
}
```

Result:
```
"fields": {
                    "foo": [
                        true
                    ]
                }
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -20,7 +20,7 @@ class Client extends ElasticaClient
     private IndexNameMapper $indexNameMapper;
 
     /**
-     * @see \JoliCode\Elastically\Factory::buildClient
+     * @see Factory::buildClient
      */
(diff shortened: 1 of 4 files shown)


damienalexandre:
Hello, I'm late of this review sorry about that.

I don't think the scripted fields should be merged with the source as a default behavior. That's too much different from the behavior of Elastica or the native Elasticsearch client and can lead to unexpected issues (we only have one DTO for pushing and reading, the scripted field would be in an invalid state on pushing?).

The proper way to use this kind of behavior in your application is to decorate the `ResultSetBuilder` with your own, in your own codebase.

Example with Symfony:

```yaml
# config/services.yaml
services:
    MyApp\ResultSetBuilder:
        # overrides the Elastically service
        # but that service is still available as ".inner"
        decorates: elastically.abstract.result_set_builder
```

PR #205, opened 2025-05-01
fix documentation transport_config

IssamRaouf:
As per the code, the transport_config option should be placed under the client section, rather than directly at the top level of the connection configuration

**JoliCode\Elastically\Bridge\Symfony\DependencyInjection\ElasticallyExtension**

`````
 if ($transportClient = $config['client']['transport_config']['http_client'] ?? null) {
            $config['client']['transport_config']['http_client'] = new Reference($transportClient);
}
`````

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,9 @@ elastically:
             client:
                 hosts:
                     - '127.0.0.1:9200'
+                # Use HttpClient component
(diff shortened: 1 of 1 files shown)


lyrixx (approved):
Good catch. Thanks

Issue #211, opened 2025-10-15
index_class_mapping and other are not available in Symfony


damienalexandre:
The factory and the symfony DIC are not building the client the same way.

Issue #191, opened 2024-09-04
[Chore] Mapping file not found


EOL-Fred:
Hi,

First, here is my configuration :
Elastically 1.9.2 (tried with 1.9.0 and 1.9.1)
Symfony 7.0.6 with PHP 8.2.21

I have a similar situation to Issue #187, when I try to create the index `land-registry` I have this error message :
```
In YamlProvider.php line 37:
Mapping file "/land-registry_mapping.yaml" not found. Please check your configuration.
```
This is the configuration files I use :
```
#config/packages/elastically.yaml
elastically:
    connections:
        default:
            client:
                host:                '%env(ELASTICSEARCH_HOST)%'
                # If you want to use the Symfony HttpClient (you MUST create this service)
                #transport:           'JoliCode\Elastically\Transport\HttpClientTransport'

            # Path to the mapping directory (in YAML)
            mapping_directory:       '%kernel.project_dir%/config/elasticsearch'

            # Size of the bulk sent to Elasticsearch (default to 100)
            bulk_size:               100

            # Mapping between an index name and a FQCN
            index_class_mapping:
                land-registry:     App\Model\Location\LandRegisty

            # Configuration for the serializer
            serializer:
                # Fill a static context
                context_mapping:
                    foo:                 bar

            # If you want to add a prefix for your index in elasticsearch (you can still call it by its base name everywhere!)
            # prefix: '%kernel.environment%'
```

```
#config/elasticsearch/land-registry_mapping.yaml
settings:
    number_of_replicas: 0
    number_of_shards: 1
    refresh_interval: 60s
mappings:
    dynamic: false
    properties:
        prefix:
            type: text
        section:
            type: text
        planNb:
            type: text
        sheet:
            type: text
        northBound:
            type: double
        southBound:
            type: double
        eastBound:
            type: double
        westBound:
            type: double
        type:
            type: keyword
        layer:
            type: keyword
        text:
            x:
                type: double
            y:
                type: double
            height:
                type: double
            text:
                type: text
            rotation:
                type: double
        polygon:
            type: geo_shape
        extended:
            IDU:
                type: text
            COAR:
                type: text
            INDP:
                type: text
            SUPF:
                type: text
```

SalvadorCardona:
I has exactly the same problem of you.

My config is good :

![image](https://github.com/user-attachments/assets/db6560af-15f4-43b4-bebc-480cfd765b2a)

When i use the debugger, i can see in Yaml Provider the configuration Directory is not set 

[...]

PR #212, opened 2025-10-20
chore(tooling): Replace Makefile with Castor and add tests for Symfony

damienalexandre:
- Replace Makefile with Castor :beaver: for better DX
- Improve the documentation
- Add tests while investigating #211 and #191

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,10 +87,10 @@ jobs:
               with:
                   php-version: ${{ matrix.php-version }}
 
-           - name: Setup castor
-             uses: castor-php/setup-castor@v0.1.0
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -86,8 +86,9 @@ jobs:
               uses: shivammathur/setup-php@v2
               with:
                   php-version: ${{ matrix.php-version }}
+                  tools: castor
```
pyrech: As you support very old PHP versions, you can also use the official action to install a static binary of Castor embedding a modern PHP version

```suggestion

           - name: Setup castor
             uses: castor-php/setup-castor@v0.1.0
```

PR #215, opened 2025-11-06
feat: allow specifying index name

### LINKED ISSUE #213: RFC: Allow specifying target index name(s) in IndexationRequest
## Summary

This RFC proposes extending the `IndexationRequest` class to allow explicitly specifying one or more target Elasticsearch index names when creating an indexation request. This enables a blue-green deployment pattern where a new index can be created, populated, and then atomically marked as live through alias swaps, without relying on automatic index alias resolution.

### Current Behavior

Currently, Elastically automatically resolves the target index based on the entity's class mapping:

```php
$bus->dispatch(new IndexationRequest(Product::class, '1234567890'));
```

The `IndexationRequestHandler` uses `IndexNameMapper::getIndexNameFromClass()` to determine which index receives the document. This works well for standard indexation workflows but has limitations:

- **Single index per class**: Documents of a given class always go to the same index
- **Alias-dependent workflow**: Requires reliance on Elasticsearch aliases to manage index versions
- **Limited blue-green patterns**: Difficult to implement workflows where you create a new index, populate it in parallel, then switch traffic via alias changes

### Proposed Use Case

In a blue-green deployment pattern for Elasticsearch:

1. Create a new index (e.g., `products_202512345`)
2. Populate it with documents while the old index (`products_202512000`) remains live
3. Once population is complete and validated, atomically swap the alias to point to `products_202512345`
4. Clean up the old index

Currently, this pattern requires either:
- Managing two separate index names in the codebase
- Creating parallel document flows
- Complex coordination outside of Elastically

The proposed solution would allow explicitly targeting specific indices during indexation:

```php
// Target the new blue-green index instead of the alias
$bus->dispatch(new IndexationRequest(Product::class, '1234567890', targetIndex: 'products_202512345'));
```

### 1. Extend `IndexationRequest` Class

Add an optional `targetIndex` parameter to `IndexationRequest`:

```php
namespace JoliCode\Elastically\Messenger;

class IndexationRequest implements IndexationRequestInterface
{
    private string $className;
    private string $id;
    private string $operation;
    private ?string $targetIndex;

    public function __construct(
        string $className,
        string $id,
        string $operation = self::OP_INDEX,
        ?string $targetIndex = null
    ) {
        $this->className = $className;
        $this->id = $id;
        $this->operation = $operation;
        $this->targetIndex = $targetIndex;
    }

    public function getClassName(): string
    {
        return $this->className;
    }

    public function getId(): string
    {
        return $this->id;
    }

    public function getOperation(): string
    {
        return $this->operation;
    }

    public function getTargetIndex(): ?string
    {
        return $this->targetIndex;
    }
}
```

### 2. Update `IndexationRequestHandler`

Modify the `schedule()` method to use the explicit target index when provided:

```php
private function schedule(Indexer $indexer, IndexationRequest $indexationRequest)
{
    // Use explicit target index if provided, otherwise fall back to class mapping
    if ($indexationRequest->getTargetIndex() !== null) {
        $indexName = $indexationRequest->getTargetIndex();
    } else {
        try {
            $indexName = $this->indexNameMapper->getIndexNameFromClass(
                $indexationRequest->getClassName()
            );
        } catch (RuntimeException $e) {
            throw new UnrecoverableMessageHandlingException(
                'Cannot guess the Index for this request. Dropping the message.',
                0,
                $e
            );
        }
    }

    // ... rest of the method
}
```

This would also be usefull for https://github.com/jolicode/elastically/issues/161.

### 3. Backward Compatibility

- The `targetIndex` parameter is optional and defaults to `null`
- When `null`, behavior is identical to current implementation
- All existing code continues to work without modification
- This is a purely additive change

Happy to contribute the changes for this / discuss this on Symfony's slack.

GaryPEGEOT:
Fixes #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,8 +420,8 @@ $bus->dispatch(new IndexationRequest(Product::class, '1234567890', targetIndex: 
 // Or 
 
 $bus->dispatch(new MultipleIndexationRequest([
-    $bus->dispatch(new IndexationRequest(Product::class, '123', targetIndex: 'custom_index_1')),
-    $bus->dispatch(new IndexationRequest(Product::class, '123', targetIndex: 'custom_index_2')),
+    new IndexationRequest(Product::class, '123', targetIndex: 'custom_index_1'),
+    new IndexationRequest(Product::class, '123', targetIndex: 'custom_index_2'),
 ]));
 
 ```


GaryPEGEOT:
> Hello 👋 😄
> 
> Looks good, tests are red as you can see.
> 
> This is a useful feature, can you add a note in the changelog, as well as an example in the readme? (in the "demo" code block https://github.com/jolicode/elastically?tab=readme-ov-file#demo).
> 
> Thanks!

I forgot to rebase, should be good now! I've also updated the doc with examples

damienalexandre:
Thanks a lot :clap:

Issue #36, opened 2020-07-31
Add phpdocumentor/reflection-docblock as a require


lamasfoker:
At [this line](https://github.com/jolicode/elastically/blob/master/src/Client.php#L108) you instantiate `\Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor`, but as the [composer.json](https://github.com/symfony/property-info/blob/master/composer.json#L45) of symfony/property-info said if you use this class you have to make phpdocumentor/reflection-docblock as a require.

damienalexandre:
Thanks @lamasfoker for this report,
as I said on https://github.com/monsieurbiz/SyliusSearchPlugin/issues/34#issuecomment-667210250 ; I think this static Serializer is a mistake and it should be a required option of the Client instead.

I will add the requirement right now, and work on that other matter later.

damienalexandre:
Released in v1.0.2, thanks!

PR #224, opened 2026-06-22
allow phpdocumentor/reflection-docblock v6

pableu:
I think we can safely allow Version 6 of phpDocumentor/ReflectionDocBlock. 

The symfony/property-info dependency will take care of the exact version, it has detailed required and conflicts. 

See also #36.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "php": ">=8.0",
         "ext-json": "*",
         "nyholm/psr7": "^1.8",
-        "phpdocumentor/reflection-docblock": "^4.4 || ^5.0",
+        "phpdocumentor/reflection-docblock": "^4.4 || ^5.0 || ^6.0",
         "ruflin/elastica": "^8.0 || ^9.0",
         "symfony/deprecation-contracts": "^2.4 || ^3.0",
         "symfony/property-access": "^5.4 || ^6.4 || ^7.0 || ^8.0",


lyrixx:
Thanks for the PR! I looked into the CI failures. There are two separate problems.

### 1. Symfony 6 jobs: Composer installs an older property-info to get docblock v6

`symfony/property-info` only declares a conflict with `reflection-docblock` v6 starting from these releases:

| Branch | Releases without the conflict | With docblock v6 |
|---|---|---|
| 6.4 | ≤ 6.4.32 | 6.4.32 throws the `LogicException` (symfony/symfony#63126) |
| 7.3 | ≤ 7.3.10 | no exception, no compatibility either |
| 7.4 | ≤ 7.4.5 | same (compatibility added in 7.4.6, symfony/symfony#63235) |
| 8.0 | ≤ 8.0.4 | same (compatibility added in 8.0.6) |
| 5.4 | all releases | never tested by Symfony |

To install docblock 6, Composer picks an older property-info release that has no conflict. I reproduced it locally with Symfony pinned to `6.4.*`:

```
phpdocumentor/reflection-docblock  6.0.3
symfony/property-info              6.4.32
→ LogicException: symfony/property-info v6 does not support phpdocumentor/reflection-docblock v6.
```

The exception comes from the `PhpDocExtractor` constructor, which `Factory::buildSerializer()` calls, so every SF 6.4 user would lose the serializer after a `composer update`.

A possible fix is to raise the minimum property-info versions to releases that declare the conflict, so Composer can't pick an older one:

```json
"symfony/property-info": "^5.4 || ~6.4.33 || ~7.3.11 || ^7.4.5 || ^8.0.5",
```

This drops 7.0 to 7.2, which are already EOL. SF 5.4 remains unprotected because property-info 5.4 never got the conflict. Its code is close to 6.4, which Symfony considers incompatible with v6. We need to decide what to do about 5.4.

It would also help to add a CI job that forces docblock v6 on SF 7.4 or 8.x, so this combination is actually tested.

### 2. Symfony 8 jobs: not related to this PR

These failures (`Framework kernel terminate resend`, `Controller with response`) also happen with docblock 5. They come from the Messenger issue fixed in #225, which isn't in this branch. On current `master` with docblock 6.0.3 and property-info 8.1.7, the full suite passes locally. A rebase should fix these jobs.

